PR #46, opened 2023-10-04
v0.0.2

jessicamcinchak:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "digital-planning-data-schemas",
-  "version": "0.0.1",
-  "description": "Digital Planning Schemas",
+  "version": "0.0.2",
```
DafyddLlyr, line 3: I was wondering about how we communicate breaking changes / patches below v1 and it turns out semver have this covered - https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

The suggestion is to start at 0.1.0 allowing 0.1.1 to be a patch/new feature and 0.2.0 to be a breaking change (as I read it).

What do you think?

jessicamcinchak: Good find - this feels more ideal, updated :heavy_check_mark:

DafyddLlyr (approved):

PR #58, opened 2023-10-20
fix: update missing or incorrect enum values

jessicamcinchak:
- Adds `other` option to `data.applicant.interest`
- Adds `proposal.photograph.evidence` to `FileType`
- Updates `property.site.locationPlan` &rarr; `property.drawing.locationPlan` in `FileType`
- Updates descriptions only for elevations & sections in `FileType`

Fixes validation errors thrown for a number of past applications checked here: https://github.com/theopensystemslab/planx-core/pull/166

### DIFF
--- a/types/schema/data/Property.ts
+++ b/types/schema/data/Property.ts
@@ -18,8 +18,8 @@ export interface UKProperty {
   localAuthorityDistrict: string[];
   type: PropertyType;
   boundary?: {
-    site?: GeoJSON;
-    area?: Area;
+    site: GeoJSON;
+    area: Area;
   };
   constraints?: {
     planning: PlanningConstraint[];
--- a/types/schema/data/Property.ts
+++ b/types/schema/data/Property.ts
@@ -18,8 +18,8 @@ export interface UKProperty {
   localAuthorityDistrict: string[];
   type: PropertyType;
   boundary?: {
-    site: GeoJSON;
-    area: Area;
+    site?: GeoJSON;
+    area?: Area;
   };
   constraints?: {
     planning: PlanningConstraint[];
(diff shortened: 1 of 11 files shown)


types/schema/data/Property.ts
```diff
@@ -18,8 +18,8 @@ export interface UKProperty {
   localAuthorityDistrict: string[];
   type: PropertyType;
   boundary?: {
-    site: GeoJSON;
-    area: Area;
+    site?: GeoJSON;
+    area?: Area;
```
DafyddLlyr, line 22: This is the only change I don't quite understand sorry.

Is it not the case that a boundary is either `undefined` if they upload a location plan, or all fields are required if they draw a boundary, i.e. `{ site: GeoJSON, area: Area }`

DafyddLlyr (approved):
One small question - I might just be missing something here?

types/schema/data/Property.ts
jessicamcinchak: That was also my impression of how it should be working, but empty object fields still seem to be coming through - so rather than just omitting the `boundary` key altogether it's coming through as `boundary: {}`, which was then incorrectly expecting nested keys to exist. 

Need to properly test / fix these cases (it's more than just boundary) in the passport mapping in planx-core first then can come back to tighten these types again.

DafyddLlyr: Yep - that's totally fair and makes sense if the plan is to come back and tighten up 👍 Maybe a `todo` here is a good shout?

jessicamcinchak: Decided to revert this before merging, think I'll have it worked out in the next planx-core change :+1:

PR #59, opened 2023-10-24
feat: add formal GeoJSON type

### LINKED ISSUE #41: Add proper GeoJSON type
There isn't a native GeoJSON format available https://github.com/ajv-validator/ajv-formats

Other options:
- https://www.npmjs.com/package/@types/geojson
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/geojson/index.d.ts

jessicamcinchak:
via https://www.npmjs.com/package/@types/geojson which is specification RFC 7946

resolves #41

### DIFF
--- a/types/enums/ProjectTypes.ts
+++ b/types/enums/ProjectTypes.ts
@@ -2,6 +2,7 @@
  * Values of `data.proposal.projectType`
  */
 export const ProjectTypes = {
+  alter: 'Alter a building',
   'alter.balcony': 'Add or alter a balcony',
   'alter.bayWindow': 'Add or remove a bay window',
   'alter.bayWindow.add': 'Add a bay window',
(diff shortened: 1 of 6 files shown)


types/enums/ProjectTypes.ts
```diff
@@ -217,6 +219,7 @@ export const ProjectTypes = {
   'new.temporaryStructure':
     'Erect a temporary structure for historic visitor attractions and listed buildings',
   'new.warehouse': 'Storage or distribution premises',
+  unit: 'Change of units',
```
jessicamcinchak, line 222: a couple more minor project type enum changes snuck in here - can cherry pick to their own branch if noisy, but they should be quite straightforward!

DafyddLlyr (approved):

Issue #51, opened 2023-10-11
How to access enum types?


jessicamcinchak:
Currently, we're encoding some very long enum types directly into the schema - see definitions for `"ProjectType"` & `"PropertyType"` as examples.

Open questions: 
- As consumers of the schema, is it useful to have these directly encoded in the single JSON Schema doc for reference / to do lookups against?
- Would it be easier to access via another query to an external service? What are central, open hosting options?
- What are maintance considerations of either approach when values & descriptions need to be updated?

jessicamcinchak:
See these examples for how we're accessing enums within Planx editor to do flow validation on publish:
- https://github.com/theopensystemslab/planx-core/pull/436
- https://github.com/theopensystemslab/planx-new/pull/3373

jessicamcinchak:
Closing this for now as this hasn't come up in external feedback & would be slightly addressed by a public validation endpoint #116

PR #60, opened 2023-10-25
chore: improve JSDoc annotations

jessicamcinchak:
Small tidy ups while doing presentation prep & comparing to OneApp

### DIFF
--- a/types/schema/data/Property.ts
+++ b/types/schema/data/Property.ts
@@ -11,7 +11,7 @@ export type Property = UKProperty | LondonProperty;
 
 /**
  * @id #UKRegion
- * @description UK regions sourced from planning.data.gov.uk/dataset/region, where "London" is a proxy for the Greater London Authority (GLA) area
+ * @description UK regions sourced from planning.data.gov.uk/dataset/region, where London is a proxy for the Greater London Authority (GLA) area
  */
 export type UKRegion =
   | 'North East'
--- a/types/schema/data/Property.ts
+++ b/types/schema/data/Property.ts
@@ -1,6 +1,7 @@
+import {GeoJSON} from 'geojson';
 import {PlanningConstraints} from '../../enums/PlanningConstraints';
 import {PropertyTypes} from '../../enums/PropertyTypes';
-import {Area, GeoJSON, URL} from '../../utils';
+import {Area, URL} from '../../utils';
 
 /**
  * @id #Property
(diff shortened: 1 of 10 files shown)


types/schema/data/Property.ts
```diff
@@ -29,9 +29,10 @@ export interface UKProperty {
 
 /**
  * @id #LondonProperty
- * @description Property details for sites within London
+ * @description Property details for sites within the Greater London Authority (GLA) area
  */
 export interface LondonProperty extends UKProperty {
+  region: 'London';
```
jessicamcinchak, line 35: this makes for a bit of a messy union type that's `'London' | string` - better suggestions for handling / communicating what I'm trying to get at here? 

should i just type as a proper enum with the 9 possible regions returned by https://www.planning.data.gov.uk/dataset/region? i don't think these should change often?

DafyddLlyr: This is a good question. There's basically two ways of doing this that I can think of which better describe the type we intend to be used here.

The first would be something like - 

```ts
type UKRegion = string extends "London" ? never : string;
```

This is pretty neat as TS "trick" but basically meaningless in terms of the schema.

I think a better approach would be to define a `UKRegion` string union type of all 9 just as you suggest, and reference the Planning Data link as the source of these 9 strings in a comment or JSDoc.

We can then end up with `Exclude<UKRegion, "London">` and `Extract<UKRegion, "London">` (this second one is also the same as the string literal type "London" but is maybe more meaningful?).

Hopefully they would compile correctly into JSON schema!

DafyddLlyr: One thing I'm super keen to discuss further with Planning Data is them being the custodian of these data dictionaries - I think this is a very meaningful and potentially quick win which may solve https://github.com/theopensystemslab/digital-planning-data-schemas/issues/51 in a satisfying way also.

DafyddLlyr (approved):

types/schema/data/Property.ts
jessicamcinchak: Updated to type out an enum with the 9 values - only glitch to the original suggestion of `Exclude`... and `Extract` is that a `LondonProperty` currently _extends_ a `UKProperty` which then means non-intersecting types. 

So I've gone with `UKProperty` having full `UKRegion` type as initial compromise here - still more precise than string!

DafyddLlyr: Ah ok - I think to make this work then we'd need an abstract (`Property`?) base interface and then extend both `UKProperty` and `LondonProperty` from this probably. 

What you have here though is a totally fair and sensible compromise though 👍

PR #62, opened 2023-10-26
fix: update `Applicant` definition

jessicamcinchak:
Emily spotted this one today while we were catching up about this work and how it relates to what editors define!

Removes `data.applicant.contact` & `data.applicant.agent.contact` and bumps keys like `name`, `email`, `phone`, `company` out to the main level consistent with Planx passport heirarchy. There wasn't any particularly good reason to wrap within `contact` in the first place. 

Once merged and published, I'll update migration guides accordingly (it's unlikely anyone has started parsing yet).

### DIFF
--- a/tests/usage.test.ts
+++ b/tests/usage.test.ts
@@ -43,7 +43,7 @@ describe("parsing using the 'ajv' library", () => {
   examplesToTest.forEach(example => {
     test(`accepts a valid example: ${example.data.application.type.description}`, () => {
       // addFormats() is required for types UUID, email, datetime etc
-      const ajv = addFormats(new Ajv());
+      const ajv = addFormats(new Ajv({allowUnionTypes: true}));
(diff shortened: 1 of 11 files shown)


tests/usage.test.ts
```diff
@@ -43,7 +43,7 @@ describe("parsing using the 'ajv' library", () => {
   examplesToTest.forEach(example => {
     test(`accepts a valid example: ${example.data.application.type.description}`, () => {
       // addFormats() is required for types UUID, email, datetime etc
-      const ajv = addFormats(new Ajv());
+      const ajv = addFormats(new Ajv({allowUnionTypes: true}));
```
jessicamcinchak, line 46: this fixes some warnings that were being thrown when testing ! see https://ajv.js.org/strict-mode.html#union-types

DafyddLlyr (approved):

PR #67, opened 2023-11-07
fix: better type for Responses `policyRef`

jessicamcinchak:
- `url` is proper `URL` type, rather than `string`
- `text` is always required if `policyRef` is available (might be case for `url` too once examples are updated? but not sure this is actually a planx content requirement!)


DafyddLlyr (approved):

DafyddLlyr:
I think that currently it's just a rich text field - so it will always have text (or be undefined), but making this a link is not enforced in the Editor.

PR #68, opened 2023-11-07
fix: simplify `Metadata` type

jessicamcinchak:
**changes:**
- break definition out into extensible interfaces `BaseMetadata` and `PlanxMetadata`; this feels easier to scale to future contributors & requirements are less coupled to Planx expectations
- remove redundant properties like `metadata.service.name` (instead inferrable from `data.application.type`)

**outstanding questions / follow ups:**
- get an example of Planning Data's organisation curie - can we better type/document expected value for this?
- are `createdAt` / `submittedAt` redundant? does the end-consumer actually care about the length of time between starting and submitting the application? should this is simply be one date that indicates when this payload is being submitted?


examples/data/ldcE.ts
```diff
@@ -1596,20 +1596,16 @@ export const validLDCE: Schema = {
...
-      source: 'PlanX',
-      id: '95f90e21-93f5-4761-90b3-815c673e041f',
-      createdAt: '2023-10-01 00:00:00',
-      submittedAt: '2023-10-02 00:00:00',
-    },
-    schema: {
-      url: 'https://theopensystemslab.github.io/digital-planning-data-schemas/v0.0.1/schema.json',
-    },
+    createdAt: '2023-10-01 00:00:00',
+    submittedAt: '2023-10-02 00:00:00',
```
jessicamcinchak, line 1607: I think this is a question we can ultimately answer now: are `createdAt` / `submittedAt` redundant? Does an end-consumer actually care about the length of time between starting and submitting the application? (We don't send this to any back office systems currently)

Should this is simply be one required date that indicates when this payload is being submitted?

DafyddLlyr: This is a very fair comment. I'd vote for simplifying this and just capturing a "submittedAt" here 👍

types/schema/Metadata.ts
```diff
@@ -1,26 +1,37 @@
...
+  id: UUID; // @todo align to DLUHC Planning Application API reference
+  createdAt: DateTime;
+  schema: URL;
+}
+
+/**
+ * @id #PlanxMetadata
+ * @description Additional metadata associated with applications submitted via PlanX
+ */
+export interface PlanxMetadata extends BaseMetadata {
```
DafyddLlyr, line 31: nit: `PlanXMetadata` may be more consistent?

DafyddLlyr (approved):

examples/data/ldcE.ts
jessicamcinchak: Updated accordingly!

Issue #53, opened 2023-10-12
Align to proposed planning register application standard


jessicamcinchak:
https://digital-land.github.io/specification/specification/planning-application/
:point_up_2: this standard is developed with a register/public access in mind - so it's uniquely capturing planning applications _after_ they've been received by an authority and intended to communicate additional details about status & decisions. 

We're describing a planning application at the point of submission _from_ a digital planning service (eg Planx) _to_ a back office system (eg Uniform), and we'll want to ensure our schema can feed directly into a register.

We'll want to align values, data formats, and possibly field naming of at least:
- `metadata.session.id` -> `reference`
- `metadata.service.owner` -> `organisation` 
- `data.application.type` -> `planning-application-type`
- WKT-format geometries (also #41)

DafyddLlyr:
I think aligning on this dataset for organisation / local authority might also be a good shout (I think this is the one BOPS said they were using internally?) https://www.planning.data.gov.uk/dataset/local-authority

jessicamcinchak:
Closing this one out as we're aligned across all Planning Data reference codes, but will ultimately skip other original suggestions on this list like migrating from GeoJSON to WKT.

PR #71, opened 2023-11-16
fix: update organisation to align to Planning Data reference code

jessicamcinchak:
Sourced from `reference` property here https://www.planning.data.gov.uk/entity.json?dataset=local-authority&limit=500

This [specification](https://digital-land.github.io/specification/specification/planning-application/) cites "reference code", but I'm honestly not exactly sure how I'd format this as a CURIE in JSON Schema - but this string should be a better alignment for now?

See #53 

I'll open a corresponding Planx PR to decide how to deal with this there: fetch & store at same point we get `local-authority-district` or store statically within `teams` db table & query with other metadata in planx-core?

### DIFF
--- a/examples/data/ldcE.ts
+++ b/examples/data/ldcE.ts
@@ -1227,7 +1227,7 @@ export const validLDCE: Schema = {
     },
   ],
   metadata: {
-    organisation: 'buckinghamshire',
+    organisation: 'BKM',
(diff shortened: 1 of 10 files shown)


DafyddLlyr (approved):
I think the idea of adding to teams as an additional column works well in terms of looking it up when required in PlanX.

PR #74, opened 2023-11-17
v0.2.0

jessicamcinchak:


CHANGELOG.md
```diff
@@ -9,6 +9,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The minor version will be incremented upon a breaking change and the patch version will be
 incremented for features.
 
+## [0.2.0] - 2023-11-17
+
+### Breaking
+- Renamed `result` to `preAssessment` and made optional, as applications from sources other than Planx are unlikely to provide this ([#66](https://github.com/theopensystemslab/digital-planning-data-schemas/pull/66))
```
jessicamcinchak, line 15: I think this is our first proper truly breaking change which bumps us up to `v0.2.0`, others agree? There's been a lot of "changing" types in these first weeks that I haven't classified as breaking yet because they began as optional, but that's not the case for `result` which was originally required.

DafyddLlyr: Yep - this makes sense to me.

It's all a little abstract until third-parties (BOPS, Uniform) are really using this but it's good to follow best practice here.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "digital-planning-data-schemas",
-  "version": "0.1.2",
+  "version": "0.1.3",
```
DafyddLlyr, line 3:
```suggestion
  "version": "0.2.0",
```

DafyddLlyr (changes requested):
One small typo / issue!

package.json
jessicamcinchak: Thank you for catching this !

DafyddLlyr (approved):

PR #75, opened 2023-11-28
fix: ensure `date-time` is ISO format

jessicamcinchak:
`ajv-formats` & `jsonschema` both permit a flexible interpretation of `date-time` that allows spaces, but we want to restrict to an ISO timestamp as flagged by BOPS.

The `iso-date-time` format won't be available until `ajv-formats` v3.0.0 is released, so I'm combining the existing `date-time` with a regex `@pattern` for now. There's a numer of existing issue threads on this topic, eg https://github.com/ajv-validator/ajv-formats/issues/55


schema/schema.json
```diff
@@ -1227,7 +1227,9 @@
       "type": "string"
     },
     "DateTime": {
+      "description": "Regex-based implementation of iso-date-time until available in ajv-formats@3.0.0",
       "format": "date-time",
+      "pattern": "^[0-9]{4}-((0[13578]|1[02])-(0[1-9]|[12][0-9]|3[01])|(0[469]|11)-(0[1-9]|[12][0-9]|30)|(02)-(0[1-9]|[12][0-9]))T(0[0-9]|1[0-9]|2[0-3]):(0[0-9]|[1-5][0-9]):(0[0-9]|[1-5][0-9])\\.[0-9]{3}Z$",
```
pixeltrix, line 1232: This regex is too strict now - according to [the RFC 3339 spec][1] the second fraction is optional and is any number of digits. Also the `T` and `Z` may be upper/lower case and timezone offsets are allowed, e.g. `2023-08-31T00:00:00+01:00`.

[1]: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

jessicamcinchak: Thanks for catching, examples updated to catch these more diverse scenarios and [regex sourced from here](https://gist.github.com/marcelotmelo/b67f58a08bee6c2468f8) now

schema/schema.json
```diff
@@ -1227,7 +1227,9 @@
       "type": "string"
     },
     "DateTime": {
+      "description": "Regex-based implementation of iso-date-time until available in ajv-formats@3.0.0",
```
DafyddLlyr, line 1230: 👍

DafyddLlyr (approved):
Thanks for picking this one up and the detailed comments. I'm not sure if we've set up dependabot in this repo yet but that's something we should do soon which will also help us spot when v3 is released.

PR #88, opened 2023-12-04
fix: adjust content based on testing feedback

jessicamcinchak:
Feedback coming out of https://github.com/theopensystemslab/planx-new/pull/2529

My first draft of this data dictionary was a bit too optimistic about which "Select the works" old flows and new triage flows are actually in use. I've gone through a number of example submissions and other flows now and filled out all possible project types. I _think_ this should be more comprehensive now :crossed_fingers:


types/enums/ProjectTypes.ts
```diff
@@ -119,36 +171,83 @@ export const ProjectTypes = {
...
+  'demolish.part.conservatory': 'Demolish part of a conservatory',
   'demolish.replace': 'Demolish a building and build homes in its place',
   extend: 'Extend a building or add an outbuilding',
+  'extend.balcony': 'Enlarge a balcony',
   'extend.basement': 'Add a basement extension',
   'extend.basement.extend': 'Enlarge a basement',
   'extend.basement.lightwell': 'Add a lightwell',
+  'extend.basement.new': 'Add a new basement extension',
   'extend.front': 'Add a front extension',
+  'extend.outbuilding':
```
jessicamcinchak, line 227: Services currently have mixed use of `extend.outbuilding` singular and `extend.outbuildings` plural. It's less than ideal to include both, but I'd rather be able to remove project types later when service content is tidied up than fail to validate all possible journeys through live services now.

types/enums/ProjectTypes.ts
```diff
@@ -73,13 +85,14 @@ export const ProjectTypes = {
   'alter.openings.add.door.side':
     'Add new doorways to the side of the building',
   'alter.openings.add.window': 'Add one or more new windows',
```
DafyddLlyr, line 87: This one is still `window` not `windows`?

types/enums/ProjectTypes.ts
DafyddLlyr: Agreed that whilst we're below v1 this is the most pragmatic approach. 

Removing any fields would be a breaking change in the schema but I think at this stage that's actually relatively insignificant as we'd be the ones affected (and the old ones wouldn't get used).

It might be worth looking into how we can decorate these field as deprecated before removing them - possibly overkill at this stage though if it proves tricky tbh!

types/enums/ProjectTypes.ts
DafyddLlyr: Just seen comment below about remaining singular and plural instances 👍

DafyddLlyr (approved):
A higher number of new additions than I was expecting - good work! 🕵️‍♀️

types/enums/ProjectTypes.ts
jessicamcinchak: This example is actually different than the singular/plural "duplicates" like "outbuilding" below - from what I found in content, it's actually always singular `window` in the parent case, and then children switch to plural `windows` :dizzy: 

`door` is always singular, so this was easy to confuse ! I'd love to do a big find&replace exercise to clean these up soon, and really pay attention to these subtle details to better align values in this particular enum.

PR #100, opened 2024-01-26
chore: extend `FileTypes` enum

jessicamcinchak:
This PR introduces a breaking change - but one that we think will be a much more meaningful and scalable naming convention going forward for communicating about uploaded documents, drawings, and the types of information contained in them.

Changes:
- Allowed file type enum values are updated based on discussions here: https://docs.google.com/spreadsheets/d/1ePihRD37-2071Wq6t2Y7QtBt7juySWuVP6SAF6T-0vo/edit#gid=842355497
- Existing example payloads are updated


jessicamcinchak:
[Update document types and add secondary schema for "files" when provided as data (eg Heritage Statements)](https://trello.com/c/kOBotqGJ/2755-update-document-types-and-add-secondary-schema-for-files-when-provided-as-data-eg-heritage-statements)

PR #107, opened 2024-02-02
feat: Add `RequestedFiles` (v2)

DafyddLlyr:
## What does this PR do?
 - Adds a list of requested files in `service.metadata.files`

### DIFF
--- a/types/schema/File.ts
+++ b/types/schema/File.ts
@@ -26,3 +26,14 @@ type FileTypeMap = {
  * @description Types of planning documents and drawings
  */
 export type FileType = FileTypeMap[keyof FileTypeMap];
+
+/**
+ * @id #FileCondition
+ */
+export type FileCondition = 'required' | 'recommended' | 'optional';
+
+/**
+ * @id #RequestedFileType
+ * @description Types of planning documents and drawings, with an attached condition as presented to the applicant
+ */
+export type RequestedFileType = FileType & {condition: FileCondition};
--- a/types/schema/Metadata.ts
+++ b/types/schema/Metadata.ts
@@ -26,14 +26,17 @@ export interface BaseMetadata {
 }
 
 /**
- * @id #RequestedFiles
- * @description File types requested by this service. Schema["files"] will be a subset of this list based on the user's journey through the service.
+ * @id #FileCondition
  */
-export interface RequestedFiles {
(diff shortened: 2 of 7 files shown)


types/schema/Metadata.ts
```diff
@@ -24,6 +25,16 @@ export interface BaseMetadata {
   schema: URL;
 }
 
+/**
+ * @id #ServiceFiles
+ * @description File types associated with this service. Schema["files"] will be a subset of this list based on the user's journey through the service.
+ */
+export interface ServiceFiles {
```
jessicamcinchak, line 32: nit: I really liked the name `RequestedFiles` from one of your earlier proposals - would that perhaps be more descriptive here as "Service" is already implied based on position in schema? Then `@description` "File types requested by this service" rather than "associated with" ?

DafyddLlyr: Yep totally agree! I've actually changed this back locally also - `RequestedFiles` is significantly clearer.

DafyddLlyr: Changes force-pushed 👍

types/schema/File.ts
```diff
@@ -26,3 +26,14 @@ type FileTypeMap = {
...
+/**
+ * @id #FileCondition
+ */
+export type FileCondition = 'required' | 'recommended' | 'optional';
+
+/**
+ * @id #RequestedFileType
+ * @description Types of planning documents and drawings, with an attached condition as presented to the applicant
+ */
+export type RequestedFileType = FileType & {condition: FileCondition};
```
DafyddLlyr, line 39: As we're relying on the enum `FileTypes` to dynamically generate the `FileType` type, we can't extend this interface - 

image

This means that the schema repeats the file list (have a look at the diff in `schema.json`).

I spent a fair bit of time trying to work around this to no success. The closest would be something like - 

```ts
export type RequestedFileType = {file: FileType} & {condition: FileCondition};
```

Both have some compromises for the consumer. I'll reach out to the BOPS team to explain this.

I think we're fine to proceed as-is, we can maybe re-evaluate how we handle these enum types at some point (see https://github.com/theopensystemslab/digital-planning-data-schemas/issues/51)

DafyddLlyr:
Closed in favour of #110

PR #110, opened 2024-02-07
feat: Add `RequestedFiles` (v1)

DafyddLlyr:
## What does this PR do?
 - Adds a list of requested files in `service.metadata.files`


examples/data/priorApproval.ts
```diff
@@ -772,6 +772,21 @@ export const validPriorApproval: Schema = {
...
+          },
+        ],
+        recommended: [],
+        optional: [
+          {
+            value: 'sitePlan.proposed',
+            description: 'Site plan - proposed',
+          },
+        ],
+      },
```
jessicamcinchak, line 789: nit: swapping the `required` & `optional` would be more realistic here until we get around to re-generating example payloads in full!

jessicamcinchak (approved):
Tidy solution :100:

PR #115, opened 2024-02-13
fix: add step to Github Action publish workflow to update version across example files

jessicamcinchak:
I think this should work, but admittedly haven't tried queuing up a release to test it ! Ideas ofr testing locally, or we just test on next release?

See #114

If this doesn't work, we can alternatively try editing `scripts/build-json-examples.ts` - but this script isn't currently aware of the current schema version (`$id`), which differs from package.json version - so felt like the more complicated route.


DafyddLlyr (approved):
Seems like a good solution to me!

Issue #117, opened 2024-02-13
Move `source` to BaseMetadata type


jessicamcinchak:

aaronwilliamsv1:
So as I understand it, source is currently a property of the extended interface PlanXMetaData:

https://github.com/theopensystemslab/digital-planning-data-schemas/blob/1fb9bf277e88c8f1c66b2812ff4e4f8eabc9bbe6/types/schema/Metadata.ts#L38-L49

And this issue is to move it from the extended interface, to the BaseMetaData interface:

https://github.com/theopensystemslab/digital-planning-data-schemas/blob/1fb9bf277e88c8f1c66b2812ff4e4f8eabc9bbe6/types/schema/Metadata.ts#L14-L26

My question would be, what should be the default for source? Should it still be 'PlanX'? What is a 'default' organisation? Should it still be string?

jessicamcinchak:
@aaronwilliamsv1 thanks for questions here, please see latest proposal #119 

Proposed changes: 
- `source` becomes required using an enum type, with a "default" state of "Any" (open to language suggestions here!)
- `organisation` remains required and any string value without a "default". As I understand it, `organisation` will always reflect `data.property.address` (required).

PR #118, opened 2024-02-13
feat: Make 'source' usable as a discriminated union

DafyddLlyr:
Fixes https://github.com/theopensystemslab/digital-planning-data-schemas/issues/117

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -1588,6 +1588,9 @@
         "schema": {
           "$ref": "#/definitions/URL"
         },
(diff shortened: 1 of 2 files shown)


schema/schema.json
```diff
@@ -1596,7 +1599,8 @@
         "organisation",
         "id",
         "submittedAt",
-        "schema"
+        "schema",
+        "source"
```
aaronwilliamsv1, line 1603: Should source be a requirement?

Issue #127, opened 2024-02-28
Schema version of generated examples in `/dist` incorrectly uses `@next`


jessicamcinchak:
Attempted to fix this here #114 & #115 but hasn't worked as expected on latest run, need to revisit 

See https://github.com/theopensystemslab/digital-planning-data-schemas/blob/dist/v0.4.0/examples/validLawfulDevelopmentCertificateExisting.json#L1901

Should: 
- Manually push correction to `/dist/v0.4.0`
- Update & test Github Actions steps again

jessicamcinchak:
Fixed as of latest release v0.4.1 !

PR #131, opened 2024-03-04
Update schema.json with latest changes

pixeltrix:


package.json
```diff
@@ -7,7 +7,7 @@
     "build": "pnpm build-schema:dist && pnpm build-json-examples",
     "build-json-examples": "pnpm ts-node scripts/build-json-examples.ts",
     "build-schema:dev": "pnpm ts-json-schema-generator --path 'types/**/*.ts' --out 'schema/schema.json' --type Schema --id @next --no-top-ref",
-    "build-schema:dist": "pnpm ts-json-schema-generator --path 'types/**/*.ts' --out 'schema/schema.json' --type Schema --id $VERSION --no-top-ref",
+    "build-schema:dist": "pnpm ts-json-schema-generator --path 'types/**/*.ts' --out 'schema/schema.json' --type Schema --id \"${VERSION:-@next}\" --no-top-ref",
```
jessicamcinchak, line 10: Makes sense - thanks for this fix !

jessicamcinchak (approved):
Thanks for this one!

We'll test this on next publish, and if it fixes #127 I'll rollback the original find&replace attempt in the Github workflow.

This PR also helped me understand your original questions and I hadn't realised the generated `@next` schema & examples on `main` had fallen out of sync with the type definitions - currently that's a bit too easy to forget to run manually when opening PRs so now proposing it's added to pre-commit hook here too which should help keep this much tidier and more aligned in future #132

PR #136, opened 2024-03-13
fix: update type definitions to reflect recent Greater London Authority changes and additions

jessicamcinchak:
Types based on https://editor.planx.uk/opensystemslab/london-data-hub

**Key changes:**
- `Property`, `Proposal` & `Application` are all union types with a base interface and Greater London Authority extension
- Removes property & proposal `details` sub-level as this doesn't actually exist in content and makes the above union type more complex than necessary/valuable
- Adds a lot of new types (we will _not_ have passport mappings for these yet in planx-core)
- Restructures existing `VehicleParking` &rarr; `Parking` (this existing passport mapping needs to be updated, then flag to August to publish new flow content)


types/schema/data/Proposal.ts
```diff
@@ -49,84 +35,117 @@ export interface BaseDetails {
...
+       */
+      swap: boolean;
+    }[];
+    protectedSpaces?: {
+      impact: 'loss' | 'gain' | 'change';
+      description: string;
+      // @todo add designation from enums/Nature
+      access: 'restricted' | 'unrestricted';
+      area: {hectares: number};
+    }[];
```
jessicamcinchak, line 102: `openSpaces` & `protectedSpaces` are examples of types of data we expect to capture in the upcoming list component. 

Eg in content these will look like `proposal.nature.openSpace.one.impact`, `proposal.nature.openSpace.two.impact` and so on. I'm intentionally omitting the "one", "two" etc from the schema types in favor of an array, we'll just handle the index text during passport mapping!

jessicamcinchak:
[Add new data points for information collected in householder & minor planning permission services](https://trello.com/c/eCJ36Y8U/2777-add-new-data-points-for-information-collected-in-householder-minor-planning-permission-services)

DafyddLlyr (approved):
Looking great!

PR #139, opened 2024-03-15
fix: update example `files` urls

jessicamcinchak:
Some example files were hitting up against our 6 month retention period and removed from S3, therefore throwing this error when BOPS tries to download if submitted: `Error: Failed to download private file: NoSuchKey: The specified key does not exist.` (https://planx.airbrake.io/projects/329753/groups/3759175601873853255)

There's a larger piece of work to update example data more robustly, but these fresh files will make existing example payloads usable in meantime!

### DIFF
--- a/examples/data/ldcE.ts
+++ b/examples/data/ldcE.ts
@@ -1178,35 +1178,20 @@ export const validLDCE: Schema = {
   ],
   files: [
     {
-      name: 'https://api.editor.planx.dev/file/private/r34zxe3x/RoaldDahlHut.jpg',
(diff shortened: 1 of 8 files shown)


DafyddLlyr (approved):
Ha! That's an interesting one I hadn't anticipated.

Agreed - this easily buys us 6 months to think about this one 👍

jessicamcinchak:
[Add example payloads for minor & major PP prototypes](https://trello.com/c/IACBcVXU/2754-add-example-payloads-for-minor-major-pp-prototypes)

PR #150, opened 2024-04-10
Split treeworksConsent into two separate types

pixeltrix:
Despite the [standard form][1] combining both types of tree works they follow different workflows - trees with a Tree Preservation Order (TPO) is an [application for consent][2] whereas trees in a Conservation Area (CA) is a [notification of intended works][3].

The process for trees with a TPO takes eight weeks and the process for trees in a CA is six weeks. Trees in a conservation area with an existing TPO follow the standard procedure.

[1]: https://assets.publishing.service.gov.uk/media/5b06c81de5274a1c5490ce6f/Form031_england_en_revised__2_.pdf
[2]: https://www.gov.uk/guidance/tree-preservation-orders-and-trees-in-conservation-areas#making-applications-tpo
[3]: https://www.gov.uk/guidance/tree-preservation-orders-and-trees-in-conservation-areas#Protecting-trees-in-conservation-areas

### DIFF
--- a/types/enums/ApplicationTypes.ts
+++ b/types/enums/ApplicationTypes.ts
@@ -112,6 +112,9 @@ export const ApplicationTypes = {
     'Outline Planning Permission - Approval of reserved matters (major)',
   'pp.outline.major.someReserved':
     'Outline Planning Permission - Consent for the principle of a project specifying some details (major)',
-  treeWorksConsent:
-    'Consent to carry out works to a tree in a Conservation Area or with a Tree Preservation Order',
+  wtt: 'Works to trees',
+  'wtt.consent':
+    'Works to trees - Consent to carry out works to a tree with a Tree Preservation Order',
+  'wtt.notice':
+    'Works to trees - Notification of proposed works to a tree in a Conservation Area',
 };
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -1564,11 +1564,47 @@
           "additionalProperties": false,
           "properties": {
             "description": {
-              "const": "Consent to carry out works to a tree in a Conservation Area or with a Tree Preservation Order",
+              "const": "Works to trees",
               "type": "string"
             },
             "value": {
-              "const": "treeWorksConsent",
+              "const": "wtt",
+              "type": "string"
+            }
+          },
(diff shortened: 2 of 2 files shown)


types/enums/ApplicationTypes.ts
```diff
@@ -98,4 +98,8 @@ export const ApplicationTypes = {
     'Outline Planning Permission - Consent for the principle of a project specifying some details',
   treeWorksConsent:
     'Consent to carry out works to a tree in a Conservation Area or with a Tree Preservation Order',
+  'treeWorksConsent.conservationArea':
+    'Notification of proposed works to a tree in a Conservation Area',
+  'treeWorksConsent.tpo':
+    'Consent to carry out works to a tree with a Tree Preservation Order',
```
jessicamcinchak, line 104: Sorry for delay in reviewing these - but had a chance to catch up with planners & service designers on our side today and learned a couple useful things: 

- Tree services are currently being developed using the proposed application type `wtt` "Works to trees"
  - This prefix is both a bit more consistent with others (eg `ldc`, `pa`, `pp`) and removes the word "Consent" from the parent application type as that's actually incorrect as this change points out!
- Rather than use suffixes `.conservationArea` & `.tpo`, they've suggested `wtt.consent` & `wtt.notice` to better reflect the application language part of the description and more flexibly adapt to a future where other types of planning constraint designations may fall within either

So, what do you think about:
```
wtt: 'Works to trees',
'wtt.consent': 'Works to trees - Consent to carry out works to a tree with a Tree Preservation Order',
'wtt.notice': 'Works to trees - Notification of proposed works to a tree in a Conservation Area',
```

Realise this is annoyingly a breaking change now instead of an additive one, but hopefully quite harmless as we aren't actually exchanging this application type yet?

pixeltrix: Yes, won't be an issue on our side - do you want me to update the PR?

jessicamcinchak: Great - if you're happy to update here directly that sounds good to me! Thanks!

jessicamcinchak (approved):

PR #164, opened 2024-05-01
fix: all `LondonProperty` fields should be optional besides "region"

jessicamcinchak:
Because the Greater London Authority (aka `london-data-hub` service) reporting requirements don't apply consistently to all application types - eg Listed Building Consent applicants will _not_ be asked about their property's title number or EPC status.

See thread https://opensystemslab.slack.com/archives/C5Q59R3HB/p1714492246639809

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -4331,11 +4331,9 @@
         }
       },
       "required": [
-        "EPC",
         "address",
         "localAuthorityDistrict",
         "region",
-        "titleNumber",
         "type"
       ],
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
It's a larger structural task, but it would be interesting to try to encode the requirements per-service with something like a discriminated union. 

I think that this might just expand the complexity to a point where things are a bit unmanageable though if more of these exceptions come in. Something to consider!

PR #167, opened 2024-05-03
chore: better document `ApplicationFee` categories and update existing examples

jessicamcinchak:
Based on output of https://editor.planx.uk/opensystemslab/fee-calculator-new

### DIFF
--- a/types/schema/data/Application.ts
+++ b/types/schema/data/Application.ts
@@ -49,27 +49,69 @@ export interface ApplicationFee {
    * @description Breakdown of calculated fee in GBP by category of development, based on the scales defined in The Town and Country Planning Regulations https://www.legislation.gov.uk/uksi/2012/2920/schedule/1/part/2
    */
   category?: {
+    /**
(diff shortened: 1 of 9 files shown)


types/schema/data/Application.ts
```diff
@@ -49,27 +49,69 @@ export interface ApplicationFee {
    * @description Breakdown of calculated fee in GBP by category of development, based on the scales defined in The Town and Country Planning Regulations https://www.legislation.gov.uk/uksi/2012/2920/schedule/1/part/2
    */
   category?: {
+    /**
+     * @description Category 1 - New homes
+     */
     one?: number;
-    oneToFour?: number;
```
jessicamcinchak, line 53: Going through the service again, I realised that these ranges (any variable with `To`) are actually booleans used for internal automations only, so they don't need to be included here !

DafyddLlyr (approved):
Really happy with the level of documentation we're getting to with the `@description` JSON Schema notation 👍

PR #168, opened 2024-05-03
fix: `Ownership` should include `interest` plus updated example payloads

jessicamcinchak:
Based on feedback from August about recent service changes to how Planx collects ownership info.

- Moves `data.applicant.interest` &rarr; `data.applicant.ownership.interest` for more meaningful heirarchy
- Updates example payloads to reflect changes to ownership service https://editor.planx.uk/opensystemslab/ownership
  - LDCs no longer set `certificate`
  - LDCs no longer report the `noticeDate` for individual owners, but simply if `noticeGiven` (boolean) and, if false,  `noNoticeReason` (string)
  - Planning Permission & LBC assign a certificate and ask more detailed questions about agricultural tenants, publishing of notice, and a declaration
  
image (8)

### DIFF
--- a/types/schema/data/Applicant.ts
+++ b/types/schema/data/Applicant.ts
@@ -63,23 +63,40 @@ export interface Ownership {
    * @description Do you know the names and addresses of all owners and agricultural tenants?
    */
   ownersKnown?: 'all' | 'some' | 'none';
-  /**
-   * @description Names and addresses of all known owners and agricultural tenants, including date of notice or reason requisite notice has not been given, if applicable
-   */
-  owners?: {
-    interest?: 'owner' | 'tenant' | 'occupier' | 'other';
-    name: string;
-    address: Address | string;
-    noticeGiven?: boolean;
-    noticeDate?: Date;
-    noNoticeReason?: string;
-  }[];
+  owners?: Owners[];
   /**
    * @description Declaration of the accuracy of the ownership certificate, including reasonable steps taken to find all owners and publish notice
    */
   declaration?: {
     accurate: true;
   };
+}
+
+/**
+ * @id #Owners
+ * @description Names and addresses of all known owners and agricultural tenants, including confirmation or date of notice, or reason requisite notice has not been given if applicable
+ */
+export type Owners = OwnersNoticeGiven | OwnersNoNoticeGiven | OwnersNoticeDate;
+
+export interface BaseOwners {
+  name: string;
+  address: Address | string;
+  interest?: 'owner' | 'tenant' | 'occupier' | 'other';
+}
+
+// LDC requires `noticeGiven`, and `noNoticeReason` if false
+export interface OwnersNoticeGiven extends BaseOwners {
+  noticeGiven: true;
+}
+
+export interface OwnersNoNoticeGiven extends BaseOwners {
+  noticeGiven: false;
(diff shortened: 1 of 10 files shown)


types/schema/data/Applicant.ts
```diff
@@ -68,18 +64,21 @@ export interface Ownership {
...
-   * @description Names and addresses of all known owners and agricultural tenants
+   * @description Names and addresses of all known owners and agricultural tenants, including date of notice or reason requisite notice has not been given, if applicable
    */
   owners?: {
+    interest?: 'owner' | 'tenant' | 'occupier' | 'other';
     name: string;
     address: Address | string;
+    noticeGiven?: boolean;
     noticeDate?: Date;
+    noNoticeReason?: string;
```
DafyddLlyr, line 75: From the PR description it sounds like it should be that if `noticeGiven === true`, we don't need a `noNoticeReason`, and if `noticeGiven === false`, then `noNoticeReason` is required.

We can achieve this with a discriminated union of these types which I _think_ more accurately describes what's expected.

JSON Schema then handles this by using an `anyOf`

```ts
// Owners is one or the other
type Owners = OwnersNoticeGiven | OwnersNoNoticeGiven;

// Put the shared properties in one place
interface BaseOwners { 
  interest?: 'owner' | 'tenant' | 'occupier' | 'other';
  name: string;
  address: Address | string;
}

// Extend with the "discriminating" key
interface OwnersNoticeGiven extends BaseOwners {
  noticeGiven: true;
}

// Use difference in discriminator key `noticeGiven` to add additional properties
interface OwnersNoNoticeGiven extends BaseOwners {
  noticeGiven: false;
  noNoticeReason: string;
}

// Valid examples
const testNoticeNotGiven: Owners = {
  interest: 'owner',
  name: 'my name',
  address: 'my address',
  noticeGiven: false,
  noNoticeReason: 'some reason',
};

const testNoticeGiven: Owners = {
  interest: 'owner',
  name: 'my name',
  address: 'my address',
  noticeGiven: true,
};
```

types/schema/data/Applicant.ts
```diff
@@ -68,18 +64,21 @@ export interface Ownership {
...
   /**
-   * @description Names and addresses of all known owners and agricultural tenants
+   * @description Names and addresses of all known owners and agricultural tenants, including date of notice or reason requisite notice has not been given, if applicable
    */
   owners?: {
+    interest?: 'owner' | 'tenant' | 'occupier' | 'other';
     name: string;
     address: Address | string;
+    noticeGiven?: boolean;
     noticeDate?: Date;
```
DafyddLlyr, line 74: Is this property still required?

> LDCs no longer report the `noticeDate` for individual owners

DafyddLlyr (approved):
One question, one suggestion. Both non-blocking!

types/schema/data/Applicant.ts
jessicamcinchak: Yep - this is a smart improvement & makes expectactions a bit more clear here - updated!

types/schema/data/Applicant.ts
jessicamcinchak: Correct - it's actually a valid third scenario in the discriminated union above for PP & LBC apps :white_check_mark:

PR #178, opened 2024-06-12
feat: add enums related to Minor PP applications and GLA reporting for residential units

jessicamcinchak:
Reflect values that will tracked via new List component schemas introduced here: https://github.com/theopensystemslab/planx-new/tree/main/editor.planx.uk/src/%40planx/components/List/schemas


jessicamcinchak:

types/enums/DevelopmentTypes.ts
```diff
@@ -0,0 +1,9 @@
+export const DevelopmentTypes = {
+  changeOfUse: 'Change of use',
+  changeOfUseFrom: 'Change of use of existing single home',
```
ianjon3s, line 3: Should probably be `Change of use of an existing single home` to be consistent with `Change of use to a home`

types/enums/UseClasses.ts
```diff
@@ -0,0 +1,14 @@
+export const GLAUseClasses = {
+  bTwo: 'B2 - General industry',
+  bEight: 'B8 - Storage and distribution',
+  cOne: 'C1 - Hotels',
+  cTwo: 'C2 - Residential institutions',
+  cTwoA: 'C2a - Secure residential institutions',
+  cThree: 'C3 - Dwellinghouses',
+  cFour: 'C4 - Houses in multiple occuplation (HMOs)',
```
ianjon3s, line 8: Spelling mistake: `occuplation` -> `occupation`

ianjon3s (approved):
Couple of small spelling/grammar comments

types/enums/UseClasses.ts
jessicamcinchak: Thanks for catching these, both updated!

PR #180, opened 2024-06-12
feat: support files submitted as "data"

jessicamcinchak:
For certain supported file types, Planx is beginning to offer applicants an option to provide the necessary information by uploading a document _or_ by answering a set of structured questions equivalent to the contents of the traditional physical document.

Currently, if the user choses the question route, those answers are exclusively being passed along via the `responses` section of the schema. But we want to make them easier to parse for reporting, so this introduces a _new_ optional `data.files` key.

Based on these Planx services:
- https://editor.planx.uk/opensystemslab/design-and-access-statement-form,mQQItUM4m0
- https://editor.planx.uk/opensystemslab/heritage-statement-form,r5w5NAGZvv

### DIFF
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -462,6 +462,42 @@
           "additionalProperties": false,
           "properties": {
             "description": {
+              "const": "Consent to make small (minor material) changes to a project with planning permission",
+              "type": "string"
+            },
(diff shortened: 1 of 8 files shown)


jessicamcinchak:

PR #192, opened 2024-07-01
feat: add Residential Unit types captured via List component

jessicamcinchak:
**Based on:**
- https://editor.planx.uk/opensystemslab/residential-units
- https://github.com/theopensystemslab/planx-new/tree/main/editor.planx.uk/src/%40planx/components/List/schemas/ResidentialUnits

**Changes:**
- Adds residential unit type definitions to ODP Schema
  - Sites within the UK can have `data.property.units.residential` and/or `data.proposal.units.residential`
  - Sites within the GLA will have `data.proposal.units.residential` with possible children `new`, `rebuilt`, `removed`, `retained`
- Updates existing Prior Approval example (in GLA with 1 rebuilt res unit)
- Residential units introduce a lot of new enums - the type definitions to convert the enum to a typescript value/description pair were adding a lot of clutter to schema type files - so now they're co-located with their enum definition which I prefer (sorry this adds so much noise though!). This refactor does _not_ actually effect the generated `schema.json`, changes there only reflect _newly introduced_ types

**Next:**
- Add `units.residential` types to minor, major and other applicable examples


jessicamcinchak:

types/schema/data/Proposal.ts
```diff
@@ -104,13 +109,21 @@ export interface BaseProposal {
...
+    }[];
+  };
 }
 
 /**
  * @id #LondonProposal
  * @description Proposal details for project sites within the Greater London Authority (GLA) area
  */
-export interface LondonProposal extends BaseProposal {
+export interface LondonProposal extends Omit<BaseProposal, 'units'> {
```
jessicamcinchak, line 126: `units` has an entirely different shape depending if the site is in `London` or the wider UK, so `Omit` allows us to essentially ignore/overwrite the UK definition when extending the interface for London.

There's other similar variables we're going to need to do this for soon too like `parking`.

types/enums/PlanningConstraints.ts
```diff
@@ -31,3 +33,40 @@ export const PlanningDesignations = {
...
+type NonIntersectingPlanningDesignation = {
+  intersects: false;
+} & BasePlanningDesignation;
+
+/**
+ * @description A planning designation that does intersect with the proposed site, per the DE-9IM spatial relationship definition of intersects
+ */
+type IntersectingPlanningDesignation = {
+  intersects: true;
+  entities: Entity[] | [];
```
DafyddLlyr, line 63: What's the reason this isn't strictly an `Entity[]`?

types/schema/data/Property.ts
```diff
@@ -69,6 +71,14 @@ export interface UKProperty {
       lastUseEndDate: Date;
     };
   };
+  units?: {
+    residential: {
+      tenure: UKTenureType;
```
DafyddLlyr, line 76: Quick naming question - do we want to have "UK" / "National" just to be implicit in all these types, with subcategories based on them?

So`TenureType` is always assumed to be national unless prefixes/suffixed by a region? e.g. `GLATenureType` or `TenureTypeGLA`

Main reason to flag this would be if something is currently national (e.g. `Materials`), but later becomes more distinct for a region it would mean changing the definition of the original type + adding a new one (e.g. `MaterialsUK` + `MaterialsScotland` -  a breaking change) vs just adding a new type (`MaterialsScotland` - a minor change).

types/schema/data/Proposal.ts
DafyddLlyr: This is a neat solution and maintains the relationship between the two `Proposal` types to pick up any future changes that will affect both 👍

DafyddLlyr (approved):
Looks good - one minor language question that can be addressed here, or discussed later.

Co-locating enums and types seems like a good call 👍

types/enums/PlanningConstraints.ts
jessicamcinchak: This is a bit of a weird relic. 

You'll find it in existing examples like this (constraint values which don't map to a single Planning Data dataset etc): 
```
{
  value: 'designated',
  description: 'Designated land',
  intersects: true,
  entities: [],
},
```

I think it was a strategy to _require_ `entity` when intersects is `true`. But we could also make entity optional, then enforce `entities: Entity[]`. I'll revisit this one in another PR as a number of other examples will have to be updated!

DafyddLlyr: Thanks for the explanation 👍

types/schema/data/Property.ts
jessicamcinchak: This is a good question. In my mind, I'm using the UK prefix for specific cases when the GLA & UK categories have _no overlap_, which feels a little different than cases where the GLA categories _extend_ national categories. 

But you're right that this could very well create unnecessary breaking changes later. I'll open a follow-up re-naming PR before we do a proper release to remove the UK prefixes and make all of this consistent (it's going to touch a lot of files)!

DafyddLlyr: Yep this seems like the best approach here 👍

PR #193, opened 2024-07-02
feat: Minor Planning Permission example payload

jessicamcinchak:
Relies on #192 

Adds a Minor Planning Permission example for a site within the UK. 

Another Minor example for a GLA site coming soon in a followup PR :+1:


jessicamcinchak:

types/schema/data/Proposal.ts
```diff
@@ -109,14 +106,7 @@ export interface BaseProposal {
...
     newBuild?: {count: number};
   };
-  units?: {
-    residential: {
-      development: DevelopmentType;
-      type: UKResidentialUnitType;
-      bedrooms: number;
-      identicalUnits: number;
-    }[];
-  };
```
jessicamcinchak, line 119: I incorrectly had `development` rather than `tenure` here - which means we can actually share the same "UK Residential Unit" definition between `property` & `proposal` - hence moving this definition to `shared.ts` now.

DafyddLlyr (approved):

PR #181, opened 2024-06-26
chore: PA - larger extension example payload

DafyddLlyr:

PR #195, opened 2024-07-03
feat: Prior Approval (larger extension) example payload

DafyddLlyr:
Cherry picked from https://github.com/theopensystemslab/digital-planning-data-schemas/pull/181 as the re-organisation of the examples caused a messy rebase.

Apologies for the delay in getting back to this one!

## To do
 - [x] PR comments from #181

### DIFF
--- a/examples/data/priorApproval/largerExtension.ts
+++ b/examples/data/priorApproval/largerExtension.ts
@@ -1,0 +1,1546 @@
+import {Schema} from '../../../types/Schema';
+
+const version = process.env['VERSION'] || '@next';
+
+export const priorApprovalLargerExtension: Schema = {
+  data: {
+    user: {
+      role: 'applicant',
+    },
+    applicant: {
+      type: 'individual',
+      name: {
+        first: 'William',
+        last: 'Zane',
+      },
+      email: 'areyouon@email.org',
+      phone: {
+        primary: '01234000000',
+      },
+      address: {
+        sameAsSiteAddress: true,
+      },
+      siteContact: {
+        role: 'applicant',
+      },
+    },
+    property: {
+      address: {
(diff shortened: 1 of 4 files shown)


examples/data/priorApproval/largerExtension.ts
```diff
@@ -0,0 +1,1546 @@
...
+    },
+    proposal: {
+      projectType: [
+        {
+          value: 'extend.rear',
+          description: 'Add a rear or side extension (or conservatory)',
+        },
+      ],
+      description:
+        'A 2 storey rear extension with a roof garden and built in pizza oven',
```
DafyddLlyr, line 227: Good spot here @jessicamcinchak (re: https://github.com/theopensystemslab/digital-planning-data-schemas/pull/181#discussion_r1656649100)

I've raised the question on Slack for the content team to address - it does appear that this is not correctly set in PA flows. https://opensystemslab.slack.com/archives/C5Q59R3HB/p1720004192950049

jessicamcinchak (approved):
Thanks for this one! Overall very happy for it to merge as-is.

Double-checking the original passport, there are a number of interesting/relevant `proposal.extension` variables that we don't have typed yet which would be nice to include as structured `data` in future (rather than only within `responses` here): https://api.editor.planx.dev/admin/session/80d3c3c2-0d1c-4a79-be99-912f488c2f02/summary

examples/data/priorApproval/largerExtension.ts
jessicamcinchak: Manually setting it here is great then, I did the same across other PA examples.

jessicamcinchak:
Rebased this one, fixed conflict & going to merge ! Thanks again for queuing up

PR #203, opened 2024-07-16
feat: Support multiple schema types

DafyddLlyr:
## What does this PR do?
- Refactors repository to support multiple schemas instead of a single schema
  - Type `Schema` is now `DigitalPlanningApplication`
- Simplifies build scripts by using a shared bash script
- Simplifies example generation to walk directories to generate JSON files instead of having to manually import each one
- Tests: Walk `/examples` to import all JSON files, avoid manual imports

## Changes to folder structure

The changes made are as follows - 

- `/schema/schema.json` → `/schemas/${SCHEMA}.json`

- `/examples/someExample.json` → `examples/${SCHEMA}/someExample.json`

- `/types/Schema.ts` → `/types/${SCHEMA}.index.ts`

- `/types/utils.ts` → `/types/shared/utils.ts`
  - I think more will end up in the `/shared` folder (maybe some enums?). The main motivation for moving utils there now (before it's actually shared) was just to be sure that these types would still be accessible to `ts-json-schema-generator` when they were "outside" / "above" the scope of the type being converted. Good news is that it works!

## Tree view
```
schemas/
└── ${SCHEMA}/
    └── ${SCHEMA}.json

examples/
└── ${SCHEMA}/
    ├── data/
    │    └── example.ts
    └── example.json

types/
├── schemas/
│   └── ${SCHEMA}/
│       └── index.ts
└── shared/
    └── utils.ts
```


.eslintrc.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./node_modules/gts/",
   "rules": {
-    "node/no-unpublished-import": ["error", {
+    "n/no-unpublished-import": ["error", {
```
DafyddLlyr, line 4: Required by gts 5.3.1 - https://github.com/google/gts/releases/tag/v5.3.1

> ## Bug Fixes
> **deps:** replace dependency eslint-plugin-node with eslint-plugin-n (https://github.com/google/gts/issues/865) ([efbe3a8](https://github.com/google/gts/commit/efbe3a838f40959188190dc04b4f11f45dd1aa87))

.github/workflows/publish.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         run: |
           VERSION=v${{steps.version_check.outputs.version}} pnpm build
           mkdir -p "$GITHUB_WORKSPACE/v${{ steps.version_check.outputs.version }}"
-          mv schema/* "$GITHUB_WORKSPACE/v${{ steps.version_check.outputs.version }}"
+          mv schemas/* "$GITHUB_WORKSPACE/v${{ steps.version_check.outputs.version }}"
```
DafyddLlyr, line 41: I think this is the only change needed here!

This has the nice benefit of preserving the old URLs in the format -

`https://theopensystemslab.github.io/digital-planning-data-schemas/<VERSION>/schema.json`

and generating new URLs in the format - 

`https://theopensystemslab.github.io/digital-planning-data-schemas/<VERSION>/schemas/<SCHEMA>.json`

examples/digitalPlanningApplication/data/landDrainageConsent.ts
```diff
@@ -1,9 +1,9 @@
-import {Schema} from '../../types/Schema';
-import {BaseProposal} from '../../types/schema/data/Proposal';
+import {DigitalPlanningApplication} from '../../../types/schemas/digitalPlanningApplication';
```
DafyddLlyr, line 1: I'll have a look (in another PR) at path aliases in our `tsconfig.json` - it could be nice to import from `@types`, `@utils` etc.

scripts/build-schema.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo "Running build-schema script..."
+
+# Use environment variable VERSION, defaulting to "@next" if not set
+version="${VERSION:-@next}"
+
+echo "Version set to $version"
+
+dirs=("digitalPlanningApplication")
+types=("DigitalPlanningApplication")
```
DafyddLlyr, line 11: When we get new schemas, we'll just need to add them to this list, e.g. 

```bash
dirs=("digitalPlanningApplication", "mySchemaDirectory")
types=("DigitalPlanningApplication", "MySchemaName")
```

types/schemas/digitalPlanningApplication/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import {FilesAsData} from './data/Files';
+import {Property} from './data/Property';
+import {Proposal} from './data/Proposal';
+import {User} from './data/User';
+
+/**
+ * @title Digital Planning Application
+ * @description The root specification for a planning application in England generated by a digital planning service
+ */
+export interface DigitalPlanningApplication {
```
DafyddLlyr, line 16: Something to consider - is `DigitalPlanningApplication` actually the right name? Is the `DigitalPlanning` prefix redundant given the context, will all schemas end up with the prefix?

tests/usage.test.ts
```diff
@@ -1,81 +1,90 @@
...
+};
+
+const schemas = [
+  {
+    name: 'DigitalPlanningApplication',
+    schema: digitalPlanningApplicationSchema,
+    examples: getJSONExamples<DigitalPlanningApplication>(
+      'digitalPlanningApplication'
+    ),
+  },
```
DafyddLlyr, line 45: New schemas will also need to be added here, and then all their examples will get picked up by tests automatically 👍

tests/usage.test.ts
```diff
@@ -1,81 +1,90 @@
...
-  examplesToTest.forEach(example => {
-    test(`accepts a valid example: ${example.data.application.type.description}`, () => {
-      // addFormats() is required for types UUID, email, datetime etc
-      const ajv = addFormats(new Ajv({allowUnionTypes: true}));
-      const validate = ajv.compile(generatedSchema);
-      const isValid = validate(example);
+  describe("parsing using the 'ajv' library", () => {
+    // addFormats() is required for types UUID, email, datetime etc
+    const ajv = addFormats(new Ajv({allowUnionTypes: true}));
+    const validate = ajv.compile(schema);
```
DafyddLlyr, line 71: Moving the instantiation of `ajv` and `jsonschema` up to the `describe()` block and not repeating in it each test speeds tests up from ~55s to ~3s.

types/schemas/digitalPlanningApplication/index.ts
jessicamcinchak: Very on board with simplifying at this stage :zap: agree dropping prefix and keeping `Application` only is less redundant and going to be easier to extend (eg `PreApplication`, `Assessment`, `Consultation`, `SiteNotice` etc)

tests/usage.test.ts
jessicamcinchak: Huge ! Thanks for spotting that misconfiguration

scripts/build-json-examples.ts
```diff
@@ -1,89 +1,46 @@
...
+      await convertTypeScriptObjectToJSON(filePath, jsonExampleFilePath);
+    }
   }
 };
 
-convertTypeScriptObjectsToJSONFiles(examplesToConvert);
+(async () => {
+  await walkDirectory('./examples');
+  console.log('All example files converted to JSON');
+})();
```
jessicamcinchak, line 46: Much appreciated dev workflow improvement :+1:

jessicamcinchak (approved):
This looks great ! Reviewed as much as one can for 65+ changed files, let's get to main & fix forward anything else we spot :slightly_smiling_face:

types/schemas/digitalPlanningApplication/index.ts
DafyddLlyr: Agreed! Will pick this up in a follow up PR shortly.

DafyddLlyr:
Thanks @jessicamcinchak! Certainly a larger and noisier PR than intended, really appreciate the speedy review 🙌

PR #207, opened 2024-07-17
feat: Add `/latest` directory containing most recently published version

DafyddLlyr:
## What does this PR do?
 - On publish of a new version, also copy the schemas, examples, and types to `/latest`
 - This means we get a stable URL always pointing to the latest published version at `https://theopensystemslab.github.io/digital-planning-data-schemas/lastest/schema/${schema}.json`
 - Relies on https://github.com/theopensystemslab/digital-planning-data-schemas/pull/206

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,6 +51,10 @@ jobs:
           # Remove JSON files from source destination
           find examples -type f -name "*.json" -delete
 
+          # Copy across new version to maintain a stable reference URL at /latest
+          rm -rf latest/*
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -51,6 +51,10 @@ jobs:
           # Remove JSON files from source destination
           find examples -type f -name "*.json" -delete
 
+          # Copy across new version to maintain a stable reference URL at /latest
+          rm -rf latest/*
```
DafyddLlyr, line 55: Removal of old files is necessary as a new version won't necessarily mean just existing files changing and new ones being added - we also want to clear out files which have been deleted as part of a new release.

jessicamcinchak (approved):

PR #209, opened 2024-07-22
Update example to include designations

EGiataganas:
This has come up from POs, as every time they want to demo a planning application with designations, they have to go through the whole process. They have created a planning application with fake data and designations and requested that these data be used so that they can generate it in Swagger automatically.


jessicamcinchak (commented):
Thanks for this contribution! Could you please propose it as a _new_ payload - eg `fullHouseholderInConservationArea` rather than overwriting existing?

More variants the merrier (but don't want to eliminate any without warning in case someone esle is relying on project types or other areas of sample app data etc) !

jessicamcinchak (commented):
One more ask - can you please follow the commands here so that this PR includes _two_ file changes including the automatically generated JSON example? https://github.com/theopensystemslab/digital-planning-data-schemas?tab=readme-ov-file#adding-examples-and-testing

jessicamcinchak (approved):

PR #215, opened 2024-07-23
feat: Initial setup of a generic application type

DafyddLlyr:
## What does this PR do?
- Sets up a new schema, temporarily named "App", which models how we could make a more specific schema for various application types
- We can vary this based on either the primary or granular application type. In the example here I'm using just the primary application type (i.e. `pp` from `pp.full.minor.technicalDetails`)
- Uses dummy type and properties to illustrate how various primary application type based models could be used

## Why is this necessary?
- This will allow us to vary which fields are required, based on application type
- As the schema has grown, more and more variables have been added which are application type specific. Out of necessity, many of these variables are currently optional as they're not provided on many user journey.
- A stricter and richer schema should be both easier to work with, and more meaningful to consumers.

## Next steps...
- Test and peer-review this approach, and then work up some examples
- Expand this approach to cover the current `DigitalPlanningApplication` schema


types/schemas/digitalPlanningApplication/enums/ApplicationTypes.ts
```diff
@@ -180,6 +180,10 @@ type GenericApplicationType<TKey extends ApplicationTypeKeys> = {
   description: (typeof ApplicationTypes)[TKey];
 };
 
+type ExtractPrimaryKeys<T> = T extends `${infer K}.${string}` ? K : T;
+
+export type PrimaryApplicationTypes = ExtractPrimaryKeys<ApplicationTypeKeys>;
```
DafyddLlyr, line 185: Before hitting on this I spent a bit of time working on constructing the list of keys using string literal types which was the approach we originally discussed.

```ts
type PrimaryAmendment = 'amendment';
type SecondaryAmendment = 'minorMaterial' | 'nonMaterial';
type Amendment =
  | PrimaryAmendment
  | `${PrimaryAmendment}.${SecondaryAmendment}`;
 ```

image

This felt like a pretty nice approach and did work (see tooltip in above screenshot), but had a few shortcomings compared to the approach using `infer`.
 - It's a whole lot more verbose
 - It lends itself very well to easily setting up a large number of permutations. Our list of application types doesn't match this pattern - for example the `pp` list has many unique combinations of secondary and tertiary keys.
 - It makes the "enum" of application types harder to parse and understand (and maintain?)
 
Here's the `infer` approach in action - 
 
image

File: types/schemas/discriminated/index.ts [hunk 2]
```diff
@@ -0,0 +1,33 @@
+import {PrimaryApplicationTypes} from '../digitalPlanningApplication/enums/ApplicationTypes';
+import {User} from './User';
+
+interface GenericApplication<T extends PrimaryApplicationTypes> {
+  primaryApplicationType: T;
```
DafyddLlyr, line 5: Not totally sold on the name for this property, but it is required to be above or at an equal hierarchy level of anywhere `T` is passed in.

This means if we need variants based on `T` outside of `data.application` (I think we do?), we can't just keep this in `data.application.applicationType`.

types/schemas/discriminated/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @internal
+ * The generic base type for all applications
+ * Takes a primary and granular application type which allows child properties to differ based on these inputs
+ * Deriving `TPrimary` from `TGranular` is possible in TS, but not in a way which is currently compatible with ts-json-schema-generator
+ */
+interface ApplicationSpecification<
+  TPrimary extends PrimaryApplicationType,
+  TGranular extends ApplicationTypeKeys,
+> {
+  applicationType: TGranular;
```
DafyddLlyr, line 18: This is required to be above or at an equal hierarchy level of anywhere `TGranular` or `TPrimary` is passed in.

This means if we need variants based on `TGranular` or `TPrimary` outside of `data.application` (I think we do?), we can't just keep this in `data.application.applicationType`.

types/schemas/discriminated/ApplicationData.ts
```diff
@@ -0,0 +1,47 @@
...
+  ApplicationTypeKeys,
+  PrimaryApplicationType,
+} from '../digitalPlanningApplication/enums/ApplicationTypes';
+
+/**
+ * @internal
+ * Base type for ApplicationData. Contains all shared properties across all application types
+ */
+export interface ApplicationDataBase<TGranular extends ApplicationTypeKeys> {
+  applicationType: TGranular;
```
DafyddLlyr, line 11: `TGranular` is repeated here in order to make this change slightly less disruptive. Note that this is just the string value and not the value/description object (we could still support this!).

types/schemas/app/ApplicationData.ts
```diff
@@ -0,0 +1,47 @@
+import {
+  ApplicationTypeKeys,
+  PrimaryApplicationType,
+} from '../digitalPlanningApplication/enums/ApplicationTypes';
+
+/**
+ * @internal
```
DafyddLlyr, line 7: `@internal` is required to not expose generic types when building the JSON schema, and to force `ts-json-schema-generator` to roll this up into its nearest parent type.

Generated JSON schema without the `@internal` notation -

image

image

types/schemas/app/index.ts
```diff
@@ -0,0 +1,63 @@
...
+  'pp'
+>;
+// TODO: All granular types
+
+export type PriorApprovalApplication = ApplicationSpecification<'pa', 'pa'>;
+// TODO: All granular types
+
+export type WorksToTreesApplications = ApplicationSpecification<
+  'wtt',
+  Extract<ApplicationTypeKeys, 'wtt' | 'wtt.consent' | 'wtt.notice'>
```
DafyddLlyr, line 36: This is a really neat solution (thanks again @jessicamcinchak!) and again points towards a simpler way of encoding enums, e.g. 

```ts
type WTTApplicationType =
  /**
   * @description Something
   */
  | "wtt"
  /**
   * @description Something else
   */
  | "wtt.consent"
  /**
   * @description Something else again
   */
  | "wtt.notice";

type ApplicationType = WTTApplicationType | PPApplicationType | PAApplicationType;

export type WorksToTreesApplications = ApplicationSpecification<
  'wtt',
  WTTApplicationType
>
```

This would solve the issue of missing or mismatched types, and having to repeat them here, if they were just defined once. I'll take a closer look at this in another PR.

types/schemas/app/index.ts
```diff
@@ -0,0 +1,63 @@
...
+>;
+
+/**
+ * @title App
+ * @description
+ * (Temporary name to not clash with the existing `Application` type)
+ * The root specification for a planning application in England generated by a digital planning service
+ */
+export type App =
```
jessicamcinchak, line 45: nit: can we call this "NextApp" or "TestApp" or "PrototypeApp" or something similar that makes it clear it's work-in-progress adn not just short-hand for existing "Application" ?

jessicamcinchak (approved):
A bit hard to review this one as I think a lot of your initial comments/questions are now outdated - _but_ this looks really good and very nice to be able to play with this setup _without_ fully replacing existing `Application` schema yet - so I say ready to merge/fix forward on `main` whenever :+1:

types/schemas/app/index.ts
DafyddLlyr: Very good shout - will make this change before I merge.

PR #217, opened 2024-07-25
fix: Update `@next` GitHub action

DafyddLlyr:
This action is currently failing due to the way generating new files and checking out dist is working. Adding the -f (force) flag resolves this.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -61,10 +61,6 @@ jobs:
           git fetch origin dist
           git checkout -b dist origin/dist
 
-      - name: Cleanup old /latest folder
-        if: steps.version_check.outputs.changed == 'true'
-        run: rm -rf latest/*
-
       - name: Commit and Push Changes
         if: steps.version_check.outputs.changed == 'true'
         run: |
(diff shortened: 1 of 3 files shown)


examples/application/data/planningPermission/fullHouseholderInConservationArea.ts
```diff
@@ -1,4 +1,4 @@
-import { Application } from "../../../../types/schemas/application";
+import {Application} from '../../../../types/schemas/application';
```
DafyddLlyr, line 1: Picked up by `gts`

.github/workflows/publish.yml
```diff
@@ -61,10 +61,6 @@ jobs:
           git fetch origin dist
           git checkout -b dist origin/dist
 
-      - name: Cleanup old /latest folder
```
DafyddLlyr, line 64: Removing this as I realised it won't actually work as expected here - being called here it would remove the newly generated version.

I think this is realistically an edge case, as the main schema file which exists here to be accessed by the `/latest` URL will be stable.

jessicamcinchak (approved):

PR #218, opened 2024-07-26
feat: Migrate `ApplicationType` to alternative enum structure

DafyddLlyr:
Frustratingly JSONSchema doesn't support description annotations per-enum (which is very likely what lead us to this more complex structure we currently use).

You can see in this PR enums with `@description` notation (PA, PP) not being pulled through into the generated schema.

For WTT, I've tried adding the descriptions as JSON within the `@description` block. This works and allows us to encode this information in the schema, but it's not typesafe or particularly consumer friendly. It might be a decent compromise though? Thoughts very welcome!

Update: The above is now outdated, please see https://github.com/theopensystemslab/digital-planning-data-schemas/pull/218#issuecomment-2256270535 - I have a solution which works well, we'll go with this going forward.


types/schemas/prototypeApplication/enums/ApplicationType.ts
```diff
@@ -0,0 +1,483 @@
...
+  | 'pp.outline.major.some.waste'
+  /**
+   * @description 'Planning Permission in Principle - Consent for the principle of a project with less than 1000 square metres floor area on a site of less than 1 hectare
+   */
+  | 'pp.pip';
+
+export type RightsOfWayOrderApplicationType =
+  /**
+   * @description Rights of Way Order - Apply to move or close a path
+   */
```
DafyddLlyr, line 439: This method doesn't work (`@description` per-enum)

types/schemas/prototypeApplication/enums/ApplicationType.ts
```diff
@@ -0,0 +1,483 @@
...
+  'rightsOfWayOrder';
+
+/**
+ * @description
+ * {
+ *   "wtt": "Works to trees",
+ *   "wtt.consent": "Consent to carry out works to a tree with a Tree Preservation Order",
+ *   "wtt.notice": "Notification of proposed works to a tree in a Conservation Area"
+ * }
+ */
```
DafyddLlyr, line 449: ...but this does (JSON in `@description` of parent).

jessicamcinchak:
Have been doing some reading on enums in JSON Schema this morning & definitely seems like simple `enum` types come with the trade off of `@description` annotations - and lots of things point towards the `oneOf`/`anyOf` dict style we have now. 

But I did find one option that I quite like - this `meta:enum` key that's part of `jsonschema2md` mentioned here https://stackoverflow.com/questions/64233370/in-json-schema-how-to-define-an-enum-with-description-of-each-elements-in-the-e

Since we know some type of improved public documentation is also on our near-roadmap here, maybe this could be a two-for-one? I suspect we'd have to "allowAdditionalKeys" to include this and extend the ts to json schema generator script ourselves to pick this up (both of which may be too large of efforts to justify!), but just wanted to flag as it feels like exactly what we want !

DafyddLlyr:
Thanks @jessicamcinchak - that link was super helpful, and lead to me [this thread](https://github.com/json-schema-org/json-schema-spec/issues/57) which suggested a combination of `anyOf` and `const`.

Take a look at aa7f59a - this looks like the solution we're looking for? 🤞

jessicamcinchak (approved):
Very happy with the latest `ApplicationType` changes - let's merge this then we can work through other enums :slightly_smiling_face:

PR #223, opened 2024-07-31
chore: `PropertyType` enum

DafyddLlyr:
Converts PropertyType from the old enum format to the new format, please see https://github.com/theopensystemslab/digital-planning-data-schemas/pull/218 for context.

### DIFF
--- a/types/schemas/prototypeApplication/enums/PropertyTypes.ts
+++ b/types/schemas/prototypeApplication/enums/PropertyTypes.ts
@@ -1,0 +1,2894 @@
+/**
+ * @description Commercial
+ */
+type Commercial = 'commercial';
+
+/**
+ * @description Agricultural
+ */
+type CommercialAgriculture = 'commercial.agriculture';
+
+/**
+ * @description Farm / Non-Residential Associated Building
+ */
+type CommercialAgricultureFarm = 'commercial.agriculture.farm';
+
+/**
+ * @description Fishery
+ */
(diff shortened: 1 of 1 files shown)


types/schemas/prototypeApplication/enums/PropertyTypes.ts
```diff
@@ -0,0 +1,2894 @@
...
+  | CommercialUtilityTelecomsMast
+  | CommercialUtilityWasteManagement
+  | CommercialUtilityWater
+  | CommercialUtilityWaterPumpControl
+  | CommercialUtilityWaterPumpDistribution
+  | CommercialUtilityWaterQualityMonitoring
+  | CommercialUtilityWaterStorage
+  | CommercialUtilityWaterTreatment
+  | CommercialUtilityWaterTreatmentWaste
+  | CommercialUtilityWaterTreatmentWater
```
DafyddLlyr, line 2729: Some pretty gnarly variable names here! I've gone for just concatenating the dot separated variables as - 
 - We can still group and list alphabetically
 - There's an easy pattern to follow for new values
 - If we simplified down, we risk clashes / issues when adding new values in future
 - They're just internal to this repo and won't be referenced outside this file

jessicamcinchak: Makes sense!

jessicamcinchak (approved):

PR #229, opened 2024-08-01
chore: `PlanningDesignation` enum

DafyddLlyr:
Converts `PlanningDesignation` from the old enum format to the new format, please see https://github.com/theopensystemslab/digital-planning-data-schemas/pull/218 for context.


types/schemas/prototypeApplication/data/shared.ts
DafyddLlyr, line 1: Due to the dependency on `Entity`, I also needed to bring this file across.

For now I've commented out the dependencies of this file so I don't have to keep pulling the import thread 🧶

DafyddLlyr, line 1: This file also needed to be migrated due to the dependency on `Entity` in `PlanningDesignation.ts`

jessicamcinchak (approved):

PR #236, opened 2024-08-29
feat: write a `prototypeApplication` schema with definitions based on primary application type

jessicamcinchak:
I know this is a VERY BIG PR - but it wasn't a particularly easy one to break up, so thanks in advance for patience reviewing ! 

**Here's a summary of key changes:**
- The `prototypeApplication` schema now has "real" type definitions which are feature-parity with the original `application` schema - but now enhanced with rules-based variations based on the primary `applicationType`
  - I've added comments below to highlight specific examples that demonstrate the new benefits of this, and places where there's still a few open questions
  - In annotating these types, I've dropped a lot of `@id` notations as we used to use per Idox feedback about this and opted for just `@descriptions`, so please take a look at the generated JSON too and see how this compares
- All `application` examples now have an equivalent `prototypeApplication` example for side-by-side comparisons of what's changing (13 example payloads in total => 26 new files here)
  
**What I haven't done here:**
- Regenerated any example application data (it'd be great to do this in the near future in order to pick up & check against latest service changes, but I don't want to muddle the side-by-side comparisons for initial feedback)
- Perfected the rules per primary application type - this is especially lacking across `data.proposal` for this MVP, definitely one to revisit separately soon ! Also, it'll never be perfect
- Defined _all_ application types - I've gone for feature-parity with types covered by the original `application` schema here, which is already "householder" centric in-line with upcoming integration expectations


types/schemas/prototypeApplication/data/Applicant.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export interface OwnersNoNoticeGiven extends BaseOwners {
+  noticeGiven: false;
+  noNoticeReason: string;
+}
+
+export interface OwnersNoticeDate extends BaseOwners {
+  noticeDate: Date;
+}
```
jessicamcinchak, line 89: `Ownership` is a nice migration story:

Our original `application` types used code comments to note which shape of owners we expected for various application types, ultimately justifying a type that was too-generic and made _all_ of these options possible: https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/types/schemas/application/data/Applicant.ts#L88-L101

Now the `LDCApplicant` definition and so on specify _only_ the ownership variation it _actually_ expects!

types/schemas/prototypeApplication/data/ApplicationData.ts
```diff
@@ -3,30 +3,213 @@ import {PrimaryApplicationType} from '../enums/ApplicationType';
...
 /**
  * TypeMap of PrimaryApplicationTypes to their specific ApplicationData models
  */
 export interface ApplicationDataVariants {
-  wtt: WTTApplicationData;
+  ldc: FeeCarryingApplicationData;
+  pa: FeeCarryingApplicationData;
   pp: PPApplicationData;
+  listed: NonFeeCarryingApplicationData;
+  landDrainageConsent: FeeCarryingApplicationData;
```
jessicamcinchak, line 212: Again, nice migration win in here of now explicitly defining which application types are "fee carrying" and which are not. There's a big semantic difference between `fee: 0` & `fee: "notApplicable"` that is now better communicated, no longer confusable based on a generic type !

types/schemas/prototypeApplication/enums/PlanningDesignation.ts
```diff
@@ -174,15 +174,15 @@ type BasePlanningDesignation =
  */
 type NonIntersectingPlanningDesignation = {
   intersects: false;
-} & BasePlanningDesignation;
+} & {value: BasePlanningDesignation};
```
jessicamcinchak, line 177: `PlanningDesignations` are actually an interesting case where I'm not totally convinced we want to migrate to a new `enum` structure? Will be curious to hear BOPS feedback on this.

Designations & constraints aren't the same typical `value/description` object enums like `ProjectTypes` (which make sense to simplify to string enums!) because _other_ properties are also/still present (eg `intersects`).

types/shared/Responses.ts
```diff
@@ -1,4 +1,4 @@
-import {URL} from '../../shared/utils';
+import {URL} from './utils';
 
 /**
  * @id #Responses
```
jessicamcinchak, line 4: Note that `Responses` are now shared between the `application` & `prototypeApplication`, but nearly all else is re-defined fresh.

The `shared` folder is generally still sparsely populated, this will make more sense to refactor once `prototypeApplication` _replaces_ `application` in my opinion ! It's felt much easier to just keep them totally seperate while doing this comparison/migration exercise rather than try to refactor/share more in same go.

schemas/prototypeApplication.json
```diff
@@ -3,654 +3,12507 @@
...
+      "required": [
+        "geometries",
+        "type"
+      ],
+      "type": "object"
+    },
+    "Integer": {
+      "format": "integer",
+      "type": "number"
+    },
```
jessicamcinchak, line 3230: This is a first attempt at adding a specific "Integer" type via utils defintions - but don't think this is quite what we want yet?

jessicamcinchak: Okay further investigation here: 
- our lib `ts-json-schema-generator` does NOT support `integer` types (yet?): https://github.com/vega/ts-json-schema-generator?tab=readme-ov-file#current-state
- alternatively `typescript-json-schema` does support with an annotation like this: https://github.com/YousefED/typescript-json-schema?tab=readme-ov-file#annotations (this is a good example of our goal state)

Is it insane to think about switching TS generation libs? This thread has a lot of comparison notes, opinions: https://github.com/vega/ts-json-schema-generator/issues/101 (one for a followup PR for sure!)

tests/usage.test.ts
```diff
@@ -41,11 +43,19 @@ const schemas = [
     schema: applicationSchema,
     examples: getJSONExamples<Application>('application'),
   },
+  // {
+  //   name: 'PrototypeApplication',
+  //   schema: prototypeApplicationSchema,
+  //   examples: getJSONExamples<PrototypeApplication>('prototypeApplication'),
+  // },
```
jessicamcinchak, line 50: This didn't work so smoothly & throws a number of type errors ! 

One to sort out in a follow-up PR I think, and for here I'm just relying on TS definitions & `pnpm build-json-examples` as the "validation" check (eg building the JSON example fails if type errors).

types/schemas/prototypeApplication/enums/FileType.ts
```diff
@@ -500,11 +500,12 @@ type WasteStorageDetails = 'wasteStorageDetails';
  */
 type WaterEnvironmentAssessment = 'waterEnvironmentAssessment';
 
+// @todo rename exported type when PrototypeApplication replaces Application
 /**
  * @id #FileType
  * @description Types of planning documents and drawings
  */
-export type FileType =
+export type PrototypeFileType =
```
jessicamcinchak, line 508: A few temporary "renames" required else we'll throw an error that more than one type with same name can't be exported !

examples/application/data/landDrainageConsent.ts
```diff
@@ -712,6 +712,6 @@ export const landDrainageConsent: Application = {
         payable: [],
       },
     },
-    schema: `https://theopensystemslab.github.io/digital-planning-data-schemas/${version}/schema.json`,
+    schema: `https://theopensystemslab.github.io/digital-planning-data-schemas/${version}/schemas/application.json`,
```
jessicamcinchak, line 715: When we re-organised the file structure, we missed updating existing example payload URLs - so those are handled here! Sorry it makes for another 13 x 2 files changed though :see_no_evil:

jessicamcinchak:
[Extend ODP Schema repository to support multiple schemas (Application, PreApp, Assessment, etc) and improve rules-based definitions (release v1.0.0🚀)](https://trello.com/c/LcsWNf2w/2945-extend-odp-schema-repository-to-support-multiple-schemas-application-preapp-assessment-etc-and-improve-rules-based-definitions-r)

types/schemas/prototypeApplication/data/Proposal.ts
```diff
@@ -1,9 +1,303 @@
...
+  /**
+   * @description Creating new buildings
+   */
+  newBuildings?: NewBuildingsOrStoreys;
+  /**
+   * @description Increasing the height of existing buildings
+   */
+  newStoreys?: NewBuildingsOrStoreys;
+  /**
+   * @description Project cost
```
DafyddLlyr, line 161: nit: Worth adding GBP?

types/schemas/prototypeApplication/data/Proposal.ts
```diff
@@ -1,9 +1,303 @@
...
+    /**
+     * @description Whether the proposal includes grey water re-use
+     */
+    grey: boolean;
+  };
+  /**
+   * @description Proposed energy sources
+   */
+  energy?: {
+    type: Array<'communityOwned' | 'heatPump' | 'solar'>;
```
DafyddLlyr, line 218:
```suggestion
  /**
   * @uniqueItems true
   */
    type: Array<'communityOwned' | 'heatPump' | 'solar'>;
```

I think we should be able to uses this annotation? ([docs](https://json-schema.org/understanding-json-schema/reference/array#uniqueItems))

DafyddLlyr (approved):
Huge effort! 🙌 

A few very minor comments - pretty much all typos and annotation suggestions.

All looks good to me, happy to help and fix forward some of the outstanding issues around tests and types.

types/schemas/prototypeApplication/data/Applicant.ts
```diff
@@ -0,0 +1,167 @@
...
+}[];
+
+export type LDCApplicant = ApplicantBase & {
+  /**
+   * @description Information about the propery owners, if different than the applicant
+   */
+  ownership:
+    | {interest: Extract<OwnershipInterest, 'owner'>}
+    | {
+        interest: OwnershipInterest; // `Exclude<OwnershipInterest, "owner">` ? But I think you can be co owner & report other owners?
```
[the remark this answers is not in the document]
jessicamcinchak: A couple of the new `prototypeApplication` example payloads have `ownership` commented out until we resolve this type - it's a good proxy for TODO ! :+1:

types/schemas/prototypeApplication/data/Property.ts
```diff
@@ -0,0 +1,147 @@
...
+} from '../enums/PlanningDesignation';
+import {PropertyType} from '../enums/PropertyTypes';
+
+export type PropertyBase = UKProperty | LondonProperty;
+
+/**
+ * @id #UKRegion
+ * @description The UK region that contains this address sourced from planning.data.gov.uk/dataset/region, where London is a proxy for the Greater London Authority (GLA) area
+ */
+export type UKRegion =
```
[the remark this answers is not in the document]
jessicamcinchak: Aha this is a great shout - I think we should probably be using "England" throughout for clarity. Going to find + replace this one in a followup PR :+1:

PR #237, opened 2024-09-19
feat: define optional `overrides` on PlanX metadata

jessicamcinchak:
PlanX fetches & suggestions administrative data to users throughout a service (eg planning constraints), and we always present users with an option to contest or change this data if it is outdated or not applicable. 

If the user has chosen to override the suggested data, we want to ensure we still send planning officers both the original "source" data and the "user" override information for assessment and feedback.

This mocks up a proposed structure for sharing this !

Will wait for #236 to merge first, then rebase here.

### DIFF
--- a/schemas/application.json
+++ b/schemas/application.json
@@ -8577,6 +8577,9 @@
             "flowId": {
               "$ref": "#/definitions/UUID"
             },
+            "overrides": {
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
Closing in favor of #240

PR #240, opened 2024-09-19
feat: define optional `overrides` on PlanX service metadata

jessicamcinchak:
PlanX fetches & suggestions administrative data to users throughout a service (eg planning constraints), and we always present users with an option to contest or change this data if it is outdated or not applicable.

If the user has chosen to override the suggested data, we want to ensure we still send planning officers both the original "source" data and the "user" override information for assessment and feedback.

This mocks up a proposed structure for sharing this !

### DIFF
--- a/types/schemas/prototypeApplication/Metadata.ts
+++ b/types/schemas/prototypeApplication/Metadata.ts
@@ -1,0 +1,49 @@
+import {
+  BaseMetadata,
+  FeeExplanation,
+  FeeExplanationNotApplicable,
+} from '../../shared/Metadata';
+import {URL, UUID} from '../../shared/utils';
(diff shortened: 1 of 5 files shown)


types/schemas/prototypeApplication/Metadata.ts
```diff
@@ -0,0 +1,49 @@
...
+  property?: {
+    type?: {
+      sourceType: string;
+      userType: string;
+    };
+    planning?: {
+      designations: {
+        value: BasePlanningDesignation;
+        sourceIntersects: true;
+        userIntersects: false;
```
DafyddLlyr, line 33: Neat and concise 👌

DafyddLlyr (approved):

PR #254, opened 2024-10-09
feat: basic schema viewer able to be deployed

jamdelion:


types/schemas/prototypeApplication/data/Proposal.ts
```diff
@@ -309,7 +309,7 @@ interface GLAGainedUnit extends GLALostUnit {
 /**
  * TypeMap of PrimaryApplicationTypes to their specific Proposal models
  */
-type ProposalVariants = {};
+type ProposalVariants = object;
```
jamdelion, line 312: Another eslint complaint:

```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.
```

types/schemas/prototypeApplication/enums/PropertyTypes.ts
```diff
@@ -2729,6 +2729,7 @@ export type PropertyType =
   | CommercialUtilityWaterTreatmentWater
   | CommercialUtilityWaterWaste
   | DualUse
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
jamdelion, line 2732: ESLint was complaining about (I think) the keyword 'object' even though `Object` is defined as its own custom type further up the file.

docs/src/App.tsx
```diff
@@ -0,0 +1,26 @@
...
+import {prototypeApplication} from 'schemas';
+
+const App = () => {
+  injectStyles();
+
+  return (
+    <MosaicProvider>
+      <Box mx="auto" py={20} px={8} style={{maxWidth: 800}}>
+        <JsonSchemaViewer
+          // name="Digital planning data schemas"
```
jamdelion, line 13: Attributes commented out because causing type errors

PR #256, opened 2024-10-15
fix: change applicant ownership tenant to lessee

augustlindemer:
Schema change to align with content. The correct term for a leaseholder with interest in land is lessee, not tenant. Agricultural tenants are a wholly separate issue and not affected by this fix.

### DIFF
--- a/schemas/application.json
+++ b/schemas/application.json
@@ -8359,7 +8359,7 @@
         "interest": {
           "enum": [
             "owner",
-            "tenant",
+            "lessee",
             "occupier",
             "other"
           ],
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Thanks @augustlindemer & welcome to schemas land :slightly_smiling_face: 

Jumped on branch with two small minor additions here:
- We're temporarily maintaining a `prototypeSchema`, in addition to orginal `application`, which is the upcoming v1 "next" preview of sorts - so I've made sure to update language there too (currently means lots of duplicated definitions we need to clean up)
- The schema repo relies on a commit hook to also automatically generate the `.json` files from the TypeScript which isn't triggered when editing via the web, but is when commiting locally via GH Desktop or CLI

PR #265, opened 2024-10-18
chore: Update lockfile, roll back TS version

DafyddLlyr:
It turns out that https://github.com/theopensystemslab/digital-planning-data-schemas/pull/253 was a breaking change - it relies on a particular version of node (or node types). We have the same issue [here](https://github.com/theopensystemslab/planx-core/pull/530) in planx-core which I spent a bit of time on but didn't quite get to the bottom of.

I think the answer is an update to Node v20 or v22 (due anyway!) across PlanX repos 👍 

On this branch I can run the pre-commit hook without issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "gts": "^5.3.1",
     "husky": "^8.0.0",
     "jsonschema": "^1.4.1",
-    "typescript": "~5.6.2",
+    "typescript": "~5.5.4",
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Aha thank you !! I ended up on a related issues thread yesterday and did try bumping `@types/node` rather than TS directly - appeciate your fresh eyes here!

PR #271, opened 2024-11-07
v0.7.1

jessicamcinchak:
A small release for exclusively the `application.json` schema.

Build scripts have been adjusted so that exclusively the `application.json` generated schema will be published to the `/dist` branch and `prototypeApplication.json` will be omitted (although it's types and examples will still be there, but there isn't an easy way to omit these without fully deleting those subdirs tempoarily - but I'm okay with that as long as actual json is omitted/not hosted yet)


scripts/build-schema.sh
```diff
@@ -7,8 +7,10 @@ version="${VERSION:-@next}"
 
 echo "Version set to $version"
 
-dirs=("application" "prototypeApplication")
-types=("Application" "PrototypeApplication")
+# dirs=("application" "prototypeApplication")
```
RODO94, line 10: This looks right to me according to what we chatted about on the dev call yesterday regarding commenting out the stuff we don't want to build yet

DafyddLlyr (approved):
Yep, this was the change I had in mind here!

PR #281, opened 2024-12-06
Add data.application.type for PreApplication schema

pixeltrix:
This mirrors the Application schema and reduces the need for branching when submitting to an endpoint that handles both pre-application and application requests.

In BOPS we're planning to create pre-applications as a type of planning application as the workflow closely mirrors that of an application and by having `applicationType` at the root it makes it difficult to handle cleanly. I've left `applicationType` in the root so that the change wouldn't break anyone depending on that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .vscode
+build
 node_modules
--- a/scripts/build-json-examples.ts
+++ b/scripts/build-json-examples.ts
@@ -33,14 +33,14 @@ const walkDirectory = async (dir: string) => {
       // Write file to mirrored directory, outside the /data folder where the TS examples are stored
       const jsonExampleFilePath = path.join(
         dir.replace('/data', ''),
-        `${path.basename(file, '.ts')}.json`
+        `${path.basename(file, '.ts')}.json`,
       );
       await convertTypeScriptObjectToJSON(filePath, jsonExampleFilePath);
     }
   }
 };
 
-(async () => {
+void (async () => {
   await walkDirectory('./examples');
   console.log('All example files converted to JSON');
 })();
--- a/scripts/build-schema.sh
+++ b/scripts/build-schema.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e
+set -o pipefail
+
 echo "Running build-schema script..."
 
 # Use environment variable VERSION, defaulting to "@next" if not set
--- a/tests/usage.test.ts
+++ b/tests/usage.test.ts
@@ -6,9 +6,11 @@ import addFormats from 'ajv-formats';
 import {Schema, Validator} from 'jsonschema';
 import {describe, expect, test} from 'vitest';
 import applicationSchema from '../schemas/application.json';
-import prototypeApplicationSchema from '../schemas/prototypeApplication.json';
+import preApplicationSchema from '../schemas/preApplication.json';
+// import prototypeApplicationSchema from '../schemas/prototypeApplication.json';
 import {Application} from '../types/schemas/application';
-import {PrototypeApplication} from '../types/schemas/prototypeApplication';
+import {PreApplication} from '../types/schemas/preApplication';
+// import {PrototypeApplication} from '../types/schemas/prototypeApplication';
 
 /**
  * Helper function to walk /examples directory and collect generated JSON files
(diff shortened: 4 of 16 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .vscode
-node_modules
\ No newline at end of file
+build
```
pixeltrix, line 2: The `tsconfig.json` file specifies `build` for the `outDir` option. The only thing in there is `tsconfig.tsbuildinfo` but just to be safe I've added it to the list of ignored files.

scripts/build-json-examples.ts
```diff
@@ -33,14 +33,14 @@ const walkDirectory = async (dir: string) => {
       // Write file to mirrored directory, outside the /data folder where the TS examples are stored
       const jsonExampleFilePath = path.join(
         dir.replace('/data', ''),
-        `${path.basename(file, '.ts')}.json`
+        `${path.basename(file, '.ts')}.json`,
```
pixeltrix, line 36: Added by the linter

scripts/build-json-examples.ts
```diff
@@ -33,14 +33,14 @@ const walkDirectory = async (dir: string) => {
...
       );
       await convertTypeScriptObjectToJSON(filePath, jsonExampleFilePath);
     }
   }
 };
 
-(async () => {
+void (async () => {
```
pixeltrix, line 43: The linter [was complaining about this][1] so i added `void` to keep it quiet

[1]: https://typescript-eslint.io/rules/no-floating-promises/#ignorevoid

scripts/build-schema.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e
+set -o pipefail
+
```
pixeltrix, line 5: This so the script fails if a subprocess fails - the pre-application schema was failing but it continued on and there was just a single 'Error' line so it was hard to spot in the output.

tests/usage.test.ts
```diff
@@ -53,46 +60,49 @@ const schemas = [
 describe.each(schemas)('$name', ({schema, examples}) => {
   const validator = new Validator();
 
-  const _applicationTypeProperty =
-    '$data.application.type.description' || '$applicationType';
-
```
pixeltrix, line 58: Removed as we don't need this now it's in the same place

tests/usage.test.ts
```diff
@@ -53,46 +60,49 @@ const schemas = [
...
-      });
-
-      test('rejects an invalid example', () => {
-        const invalidExample = {foo: 'bar'};
-        const result = validator.validate(invalidExample, schema as Schema);
-
-        expect(result.errors).not.toHaveLength(0);
-      });
-    });
+    describe.each<Application | PreApplication>(examples)(
```
pixeltrix, line 64: Jest couldn't infer the type so we have to be explicit

tsconfig.json
```diff
@@ -4,10 +4,14 @@
     "rootDir": ".",
     "outDir": "build",
     "skipLibCheck": true,
+    "esModuleInterop": true,
+    "resolveJsonModule": true
```
pixeltrix, line 8: This is so we can use `import` with the JSON files in the tests

types/schemas/application/enums/PlanningConstraints.ts
```diff
@@ -1,4 +1,4 @@
-import {Entity} from '../../../shared/Constraints';
+import {Entity} from '../../../shared/Entity';
```
pixeltrix, line 1: Entity was being defined twice so I moved it to the shared directory

types/schemas/preApplication/enums/PreApplicationTypes.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Values of `data.application.type`
+ */
+export const PreApplicationTypes = {
+  preApp: 'Pre-application advice',
+  'preApp.householder': 'Pre-application advice - Householder',
+  'preApp.minor': 'Pre-application advice - Minor',
+  'preApp.major': 'Pre-application advice - Major',
```
pixeltrix, line 8: We're just using the one type for now (`preApp`) but it's possible there may be others later so this is more an indication of intent

pixeltrix:
Closing in favour of #290

PR #289, opened 2025-01-15
chore: update `PreApplication` enums to objects like `Application`

jessicamcinchak:
The `PreApplication` schema was originally developed using string enums consistent with the `PrototypeApplication` schema (see discussions in ODP #product-devs & #odp-schemas). 

Until we make a final decision on enum formats, we'll match `PreApplication` to `Application` for consistent data exchange with BOPS.


types/schemas/prototypeApplication/data/shared.ts
```diff
@@ -2,7 +2,7 @@ import {URL} from '../../../shared/utils';
 import {UKResidentialUnitType} from '../enums/ResidentialUnitType';
 import {UKTenureType} from '../enums/TenureType';
 
-export type Entity = {
+export type PrototypeEntity = {
```
jessicamcinchak, line 5: This re-naming fixes a duplicate type export error when running `pnpm build-schema`

DafyddLlyr: Yep this was one of the build issues I also hit!

DafyddLlyr (approved):

PR #287, opened 2025-01-14
chore: update `PropertyTypes` enum values

jessicamcinchak:
Reflects latest PlanX services work on property type <> property use classes https://docs.google.com/spreadsheets/d/1P_MjwuJlTshSsz1-9yLjBtZOqim93y31AGglturIaJA/edit?gid=1046274676#gid=1046274676

Intentionally does not update the `PropertyTypes` enum for the `prototypeApplication` as this is not exchanged and it's string enum format is undecided. The `PropertyTypes` object enum is shared by both the `application` & `preApplication`

PR #293, opened 2025-01-17
fix: correct `pnpm build-schema` from main

jessicamcinchak:
Realising that when I "accepted" the typo suggestions via the GitHub UI in #287, the `pnpm build-schema` pre-commit hook wouldn't have been triggered and I didn't manually make any further commits, so I failed to pick these typo corrections up in the actual built files ! Lesson learned !

Screenshot from 2025-01-17 14-22-45

Will open a separate PR to correct files on the `/dist` branch for v0.7.2 as well.

### DIFF
--- a/schemas/application.json
+++ b/schemas/application.json
@@ -17884,7 +17884,7 @@
           "additionalProperties": false,
           "properties": {
             "description": {
-              "const": "Ordance manufacturing & processing",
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Ah good catch! Also didn't anticipate this one.

PR #298, opened 2025-01-24
chore: adjust optional preApp info for applications

augustlindemer:
New content only requires a reference (`application.preAppAdvice.reference`) if `application.preAppAdvice` is `true`.


jessicamcinchak (approved):
Thanks for this !

PR #300, opened 2025-01-27
chore: refactor more `/shared` type definitions

jessicamcinchak:
Only code re-organisations here, no functional changes! All examples still validate and tests pass. 

**Key changes:**
- Add `/shared` types for `Parking`, `MaintenanceContacts`, and `LinkedApplications`
- Update a number of `@title` annotations to drop `#` this is not required and a remnant of copying past `@id` (which do require `#`) to now `@title`

### DIFF
--- a/schemas/application.json
+++ b/schemas/application.json
@@ -5,7 +5,7 @@
   "definitions": {
     "Address": {
       "additionalProperties": false,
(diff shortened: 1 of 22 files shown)


jamdelion (approved):
Looking nice and clean :)

PR #305, opened 2025-02-03
Add Works to trees application schema

christinelangston:
Edits to specify data schema for Works to Trees (WTT) application, which has been previously unspecified. 

- Added application type in prototypeApplication/index.ts, edited Property, Proposal, Application, and Applicant typescript files to account for WTT.
- Uploaded example .ts and .json files


examples/prototypeApplication/data/worksToTrees.ts
```diff
@@ -0,0 +1,494 @@
...
+      },
+      declaration: {
+        accurate: true,
+        connection: {
+          value: 'none',
+        },
+      },
+    },
+    proposal: {
+      projectType: [],
```
jessicamcinchak, line 181: nit: for "empty" fields like `projectType` and `date` below - I think I'd expect them to be populated or fully omitted from this example ! (the "empty" values are a known payload generation quirk that we want to tidy up, but we can be more optimistic in our examples here in the meantime!)

types/schemas/prototypeApplication/index.ts
```diff
@@ -49,10 +50,17 @@ export type LandDrainageConsent = ApplicationSpecification<
...
  * @description The (prototype) root specification for a planning application in England generated by a digital planning service
  */
-export type PrototypeApplication = LDC | PA | PP | Listed | LandDrainageConsent;
+export type PrototypeApplication =
+  | LDC
+  | PA
+  | PP
+  | Listed
+  | LandDrainageConsent
+  | WTT;
```
jessicamcinchak, line 66: Once merged to `main`, we should automatically see the new `WTT` value as an available option in the dropdown selection here :star: 
https://theopensystemslab.github.io/digital-planning-data-schemas-docs/
Screenshot from 2025-02-04 17-29-51

jessicamcinchak (approved):
Welcome to schema contribution land !

All looks great, one minor non-blocking comment about a couple "empty" values in the example - we can either tidy these up here or revisit them in future/fix-forward as we know WTT is a still work-in-progress service anyways :slightly_smiling_face:

examples/prototypeApplication/data/worksToTrees.ts
christinelangston: Thanks for pointing this out, makes sense. I just committed a change to remove these empty values in the example and also adjusted the WTTProposal definition to no longer require those keys.

PR #308, opened 2025-02-17
Add Hedgerow schema

christinelangston:
Several edits to add data schema for Hedgerow Removal Notice: Added HedgerowRemovalNotice Applicant, Application Data, Proposal, and Property.
Additional edits made to allow for this:
- Made type optional in BaseApplicant (Applicant.ts file)
- Refactor Proposal by redefining ProposalBase


types/schemas/prototypeApplication/data/Proposal.ts
```diff
@@ -291,13 +294,22 @@ interface GLAGainedUnit extends GLALostUnit {
   garden: boolean;
 }
 
-export type WTTProposal = Pick<EnglandProposal, 'description' | 'boundary'>;
+export interface hedgerowRemovalNoticeProposal extends ProposalBase {
+  reason: string;
+  hedgerowLength: {metres: number};
+  hedgerowAgeLessThanThirty: boolean;
+}
```
jessicamcinchak, line 301: If we actually want to fully omit `description`, I think we can alternatively do something like this:
```
export type HedgerowRemovalNoticeProposal = Pick<ProposalBase, "boundary"> & { 
  reason: string; 
  hedgerowLength: { metres: number };
  hedgerowAgeLessThanThirty: boolean;
}
```

Which rather than _extending_ the whole ProposalBase will just pick the single optional boundary field and add it to the rest of our custom proposal definition !

examples/prototypeApplication/data/hedgerowRemovalNotice.ts
```diff
@@ -0,0 +1,346 @@
...
+          properties: {
+            label: '1',
+            'area.hectares': 0.252636,
+            'area.squareMetres': 2526.36,
+            planx_user_action: 'Drew a custom boundary',
+          },
+        },
+        area: {
+          squareMetres: 2526.36,
+        },
```
jessicamcinchak, line 95: nit: I think our `proposal.boundary.area` always expects _both_ `squareMetres` and `hectares` properties?

jessicamcinchak (approved):
Looks great, two minor comments !

types/schemas/prototypeApplication/data/Proposal.ts
christinelangston: That makes sense to me! I would defer to you, do you want to omit description here?

jessicamcinchak: Let's omit if never actually appears in the service itself as August was clarifying !

examples/prototypeApplication/data/hedgerowRemovalNotice.ts
christinelangston: Ah yes, just confirmed hectares are optional. Which makes sense, it passed the tests I ran locally. Would you prefer to have them included?

types/schemas/prototypeApplication/data/Proposal.ts
christinelangston: Made these suggested edits to Proposal.ts

PR #311, opened 2025-02-24
feat: expand `PlanningDesignations` enum values

jessicamcinchak:
**Key changes:**
- Adds new values to `PlanningDesignations` enum to represent values PlanX _anticipates_ checking in the future based on supported Application Types and Planning Data's backlog - see more context here: https://www.notion.so/opensystemslab/4-Make-sure-your-GIS-data-is-onto-Planning-data-gov-uk-75009288d7cd4f0594a39055f0087173
  - Note that sentence casing has changed for a number of `description`s based on latest service design team feedback ([see thread here](https://opensystemslab.slack.com/archives/C089LL9LR5H/p1740128497608299)), but all `value` keys remain stable
- Refactors `types/shared/Constraints.ts` to be single source of truth for all schemas (eg no longer slightly different `prototypeApplication` definitions)
  - Removes legacy types & updates example payloads accordingly (hence so many individual file changes!)


types/shared/Constraints.ts
```diff
@@ -1,5 +1,115 @@
...
+  registeredPark: 'Historic Park or Garden',
+  'road.classified': 'Classified Road',
+  safeguardedLand: 'Safeguarded land',
+  safetyHazardArea: 'Safety hazard area',
+  sourceProtectionZone: 'Source protection zone',
+  tpo: 'Tree Preservation Order (TPO) or zone',
+  watercourse: 'Watercourse',
+  'watercourse.mainRiver': 'Main river',
+  'watercourse.other': 'Other watercourse',
+};
```
jessicamcinchak, line 80: :point_up_2: This is the key change !

examples/prototypeApplication/data/lawfulDevelopmentCertificate/existing.ts
```diff
@@ -176,54 +177,62 @@ export const lawfulDevelopmentCertificateExistingPrototype: PrototypeApplication
...
               value: 'nature.ASNW',
+              description: 'Ancient Semi-Natural Woodland (ANSW)',
               intersects: false,
             },
             {
               value: 'nature.SSSI',
-              intersects: false,
-            },
-            {
-              value: 'locallyListed',
```
jamdelion, line 210: Just checking you meant to get rid of `locallyListed` here when it still appears in some of the other files :)

examples/prototypeApplication/data/landDrainageConsent.ts
```diff
@@ -159,78 +164,97 @@ export const landDrainageConsentPrototype: PrototypeApplication = {
...
             intersects: false,
           },
           {
             value: 'listed.gradeOne',
+            description: 'Listed Building - Grade I',
             intersects: true,
           },
           {
             value: 'registeredPark',
+            description: 'Historic Park or Garden',
```
jamdelion, line 227: nit: probably doesn't matter too much so feel free to ignore but we don't seem to have very consistent casing in these descriptions! I'd probably expect them to mostly be sentence case (just capitalise the first letter of the first word unless proper nouns?). I.e. `Historic park or garden` in this example, `Classified road` further down, etc. 🤷

jamdelion (approved):
All looks sensible enough to me, just a minor gripe about casing which you can certainly ignore :)

examples/prototypeApplication/data/landDrainageConsent.ts
jessicamcinchak: Yes agree these are annoying / tedious ! Historic Park or Garden is actually a proper noun as I understand it - eg https://historicengland.org.uk/listing/what-is-designation/registered-parks-and-gardens/

I'll double check with content team if we should be using "Registered Park or Garden" to better match the data value, or if "Historic" is intentional here ?

examples/prototypeApplication/data/lawfulDevelopmentCertificate/existing.ts
jessicamcinchak: Ah yes sorry, I should have explained this one - we no longer check `locallyListed` in PlanX because while available on PD it's only actually populated for a single council and not all, so leads to "false negatives". 

These examples were generated while we were still incorrectly checking it - I started removing but didn't do this consistently enough - will update correctly throughout!

jessicamcinchak: Now consistently removed throughout all examples :heavy_check_mark:

examples/prototypeApplication/data/landDrainageConsent.ts
jessicamcinchak: Some nice suggestions from services team for slightly more consistent casing implemented via [b8a1f7c](https://github.com/theopensystemslab/digital-planning-data-schemas/pull/311/commits/b8a1f7c6c614688b426e1233b627b6f76be6a1a6) :broom:

PR #312, opened 2025-02-24
feat: extend details collected under the `CommunityInfrastructureLevy` additional information form

jessicamcinchak:
See "CIL Additional Information" tab here: https://docs.google.com/spreadsheets/d/1FgULPemnwuwysrYGEkReYFXz3n3T7W0nWziU00taZE4/edit?gid=909366809#gid=909366809

### DIFF
--- a/schemas/application.json
+++ b/schemas/application.json
@@ -2770,6 +2770,169 @@
         {
           "additionalProperties": false,
           "properties": {
+            "claim": {
+              "additionalProperties": false,
+              "properties": {
+                "annexeOrExtensionExemption": {
+                  "description": "Do you wish to claim an exemption for a residential annex or extension?",
+                  "type": "boolean"
+                },
+                "charityRelief": {
+                  "description": "Is the site owned by a charity where the development will be wholly or mainly for charitable purposes, and the development will be either occupied by or under the control of a charitable institution?",
+                  "type": "boolean"
+                },
+                "selfBuildExemption": {
+                  "description": "Do you wish to claim a self build exemption for a whole new home?",
+                  "type": "boolean"
+                },
+                "socialHousingRelief": {
+                  "description": "Does the proposed development include affordable housing which qualifies for mandatory or discretionary Social Housing relief?",
+                  "type": "boolean"
+                }
+              },
+              "required": [
+                "annexeOrExtensionExemption",
+                "selfBuildExemption",
+                "charityRelief",
+                "socialHousingRelief"
+              ],
+              "type": "object"
+            },
+            "existingBuildings": {
+              "description": "How many existing buildings on the site will be retained, demolished or partially demolished as part of the development proposed?",
+              "items": {
+                "additionalProperties": false,
+                "properties": {
+                  "area": {
+                    "additionalProperties": false,
+                    "properties": {
+                      "loss": {
+                        "$ref": "#/definitions/Area",
+                        "description": "Gross internal area to be demolished"
+                      },
+                      "retained": {
+                        "$ref": "#/definitions/Area",
+                        "description": "Gross internal area to be retained"
+                      }
+                    },
+                    "required": [
+                      "retained",
+                      "loss"
+                    ],
+                    "type": "object"
+                  },
+                  "continuousOccupation": {
+                    "description": "Was the building or part of the building occupied for its lawful use for 6 continuous months of the 36 previous months (excluding temporary permissions)?",
+                    "type": "boolean"
+                  },
+                  "description": {
+                    "additionalProperties": false,
+                    "properties": {
+                      "existing": {
+                        "description": "Description of existing building or part",
+                        "type": "string"
+                      },
+                      "proposed": {
+                        "description": "Proposed use of retained gross internal area",
+                        "type": "string"
+                      }
+                    },
+                    "required": [
+                      "existing",
+                      "proposed"
+                    ],
+                    "type": "object"
+                  },
+                  "lastOccupation": {
+                    "$ref": "#/definitions/Date",
+                    "description": "When was the building last occupied for its lawful use?"
+                  },
+                  "stillInUse": {
+                    "description": "Is the building still occupied for its lawful use?",
+                    "type": "boolean"
+                  }
+                },
+                "required": [
+                  "description",
+                  "area",
+                  "continuousOccupation",
+                  "stillInUse",
+                  "lastOccupation"
+                ],
+                "type": "object"
+              },
+              "type": "array"
+            },
+            "existingPermissionPrecedingCIL": {
+              "description": "Does the application relate to details or reserved matters on an existing permission that was granted prior to the introduction of the CIL charge in the relevant local authority area?",
+              "type": "boolean"
+            },
+            "existingPermissionReference": {
+              "description": "Planning application reference number for the existing permission",
+              "type": "string"
+            },
+            "newDwellings": {
+              "description": "Does the application include creation of one or more new dwellings (including residential annexes) either through new build or conversion (except the conversion of a single dwelling house into two or more separate dwellings with no additional gross internal area created)?",
+              "type": "boolean"
+            },
+            "newMezzanine": {
+              "description": "Will the project create a new mezzanine floor within an existing building?",
+              "items": {
+                "additionalProperties": false,
+                "properties": {
+                  "area": {
+                    "$ref": "#/definitions/Area",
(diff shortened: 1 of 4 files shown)


types/shared/CommunityInfrastructureLevy.ts
```diff
@@ -13,9 +15,151 @@ type LiableForCIL = {
...
+   */
+  existingPermissionReference?: string;
+  /**
+   * @description Does the application include creation of one or more new dwellings (including residential annexes) either through new build or conversion (except the conversion of a single dwelling house into two or more separate dwellings with no additional gross internal area created)?
+   */
+  newDwellings: boolean;
+  /**
+   * @description Does the application involve new residential development (including new dwellings, extensions, conversions/changes of use, garages, basements or any other buildings ancillary to residential use)?
+   */
+  newResidentialDevelopment: boolean;
```
augustlindemer, line 63: `newResidentialDevelopment` and `newNonResidentialDevelopment` need to be optional (since they are not asked if an application is `notLiableForCIL`. I failed to specify when information is expected in the spreadsheet, sorry about that!

schemas/prototypeApplication.json
```diff
@@ -536,28 +703,315 @@
...
+              "type": "array"
             }
           },
           "required": [
+            "claim",
+            "declaration",
+            "existingPermissionPrecedingCIL",
+            "newDwellings",
+            "newNonResidentialDevelopment",
+            "newResidentialDevelopment",
```
augustlindemer, line 1012: See earlier comment. `newResidentialDevelopment` and `newNonResidentialDevelopment` are not answered if application is `notLiableForCIL`.

types/shared/CommunityInfrastructureLevy.ts
jessicamcinchak: Easy to update, thanks for spotting this!

schemas/prototypeApplication.json
```diff
@@ -536,28 +703,315 @@
...
+                "type": "object"
+              },
+              "type": "array"
             }
           },
           "required": [
+            "claim",
+            "declaration",
+            "existingPermissionPrecedingCIL",
+            "newDwellings",
```
augustlindemer, line 1010: If we want to future proof this, `newDwellings` and `floorAreaHundredPlus` can be optional (not answered) if `s73Application` or `existingPermissionPrecedingCIL` is true.

augustlindemer: If we want to capture section _4 Liability for CIL_ questions at a different level of conditionality (since them being required only depends only on application type, not liability) `floorAreaHundredPlus` is on the same level as `newDwellings` - always asked (if not s73Application or existingPermissionPrecedingCIL) reardless of liability.

schemas/application.json
```diff
@@ -2780,28 +2947,315 @@
...
+                  "area"
+                ],
+                "type": "object"
+              },
+              "type": "array"
             }
           },
           "required": [
+            "claim",
+            "declaration",
```
augustlindemer, line 3252: Is there a potential confusion around `declaration.accurate` (something we use within PlanX) and the end-of-form _declaration_ of the CIL form? We do not currently have the latter since we are not sure we want a separate confirmation. We could capture it through a set component, concern here is that we automate something that comes through as a declaration from the user?

schemas/application.json
```diff
@@ -2780,28 +2947,315 @@
...
+            "existingPermissionPrecedingCIL",
```
augustlindemer, line 3253: Technically the form would not normally capture this if `s73Application` is true. They're all mutually exclusive, but the form captures them sequentially so it's inconsistent if they're expected to be answered at all. **However**, because this and `s73Application` is always known on the basis of application type, we can always capture values for them and send them along.

schemas/prototypeApplication.json
jessicamcinchak: Now reflected in types :+1:

schemas/prototypeApplication.json
jessicamcinchak: Now reflected in types :+1:

examples/application/data/lawfulDevelopmentCertificate/existing.ts
```diff
@@ -34,6 +34,38 @@ export const lawfulDevelopmentCertificateExisting: Application = {
...
+            area: {
+              retained: {squareMetres: 75},
+              loss: {squareMetres: 0},
+            },
+            continuousOccupation: true,
+            stillInUse: true,
+            lastOccupation: '2023-10-02',
+          },
+        ],
+      },
```
jessicamcinchak, line 68: @augustlindemer - here's example data for an LDC-E claiming CIL exemption for an annexe - does it include the info you'd expect?

examples/application/data/planningPermission/fullHouseholder.ts
```diff
@@ -34,6 +34,12 @@ export const planningPermissionFullHouseholder: Application = {
           value: 'none',
         },
       },
+      CIL: {
+        result: 'notLiable',
+        declaration: true,
+        s73Application: false,
+        existingPermissionPrecedingCIL: false,
+      },
```
jessicamcinchak, line 42: @augustlindemer - similary, here's an example payload snippet for not liable - realistic (enough) & complete?

augustlindemer: This example (and the new CIL.ts structure) look like `floorAreaHundredPlus` and `newDwellings` are now only covered under S73 and in their S73 section wording, and missing from when `s73Application` & `existingPermissionPrecedingCIL` are false and with their respective wording (as per section 4 in the form) - which would be required for both liable and not liable. Other than that it's correct!

examples/application/data/lawfulDevelopmentCertificate/existing.ts
augustlindemer: Few observations:

- It misses the proposed GIA information that we would capture outside of the CIL content but feed into it, set as equal to `application.CIL.proposed.total.area.proposed` and calculate `application.CIL.proposed.total.area.net` from. We need to map these (and other totals) in the schema, as otherwise they don't appear with legible field titles/descriptions in the payload. Other totals include: `application.CIL.proposed.total.area.existing` & `application.CIL.proposed.total.area.loss`
- I would expect a false for `application.CIL.unoccupiedBuildings`
- I would expect a false for `application.CIL.newMezzanine`
- I would expect a number of existing buildings (`existingBuildings`), a value that looks to instead be used for the list (in content just `existing`) that captures details about them (I'm perpetually confused as to how lists translate into schema, but the values we use are different from the ones shown here)

jessicamcinchak: Thanks for notes! 
- Each `proposedTotalArea` data value now added :+1: 
- I was optimistically assuming that ommision of `unoccupiedBuildings` and `newMezzanine` could indicate `false`, but explicitly added now ! will have list of `buildings` or `mezzanines` respectively if "true"
- Similarly was thinking that providing `existingBuildings` as a _list of items_ communicated the count of buidings by default, but also now explicitly added !  

How does this structure feel?

examples/application/data/planningPermission/fullHouseholder.ts
jessicamcinchak: Good spot! `floorAreaHundredPlus` and `newDwellings` added back to `BaseCIL` definition so will always be captured regardless of S73 & liability

examples/application/data/lawfulDevelopmentCertificate/existing.ts
jessicamcinchak: Thanks for notes! 
- Each `proposedTotalArea` data value now added :+1: 
- I was optimistically assuming that ommision of `unoccupiedBuildings` and `newMezzanine` could indicate `false`, but explicitly added now ! will have list of `buildings` or `mezzanines` respectively if "true"
- Similarly was thinking that providing `existingBuildings` as a _list of items_ communicated the count of buidings by default, but also now explicitly added !  

How does this structure feel?

augustlindemer: This looks good, let's test it!

Re `existingBuildings` - the issue is that the list is for buildings and **parts of buildings** (the list is also further split into occupied and unoccupied). While we could sum across both lists, we can't determine which item is a part or not and parts would combine to be one unit of a whole existing building for the `existingBuildings` input.

augustlindemer (approved):
Let's test it!

PR #313, opened 2025-02-27
chore: tweak `Ownership` definitions

jessicamcinchak:
See context here https://opensystemslab.slack.com/archives/C089LL9LR5H/p1739544939178549

Brings schema definitions into sync with new "List" component schemas in PlanX https://github.com/theopensystemslab/planx-new/pull/4248

### DIFF
--- a/examples/application/data/planningPermission/major.ts
+++ b/examples/application/data/planningPermission/major.ts
@@ -46,6 +46,7 @@ export const planningPermissionMajor: Application = {
               country: 'Greece',
               postcode: '212 00',
             },
+            interest: 'Owner',
             noticeGiven: true,
           },
         ],
--- a/examples/application/data/planningPermission/major.ts
+++ b/examples/application/data/planningPermission/major.ts
@@ -46,8 +46,7 @@ export const planningPermissionMajor: Application = {
               country: 'Greece',
               postcode: '212 00',
             },
-            interest: 'Owner',
-            noticeGiven: true,
+            noticeDate: '2024-01-01',
           },
         ],
         declaration: {
--- a/types/schemas/prototypeApplication/data/Applicant.ts
+++ b/types/schemas/prototypeApplication/data/Applicant.ts
@@ -40,7 +40,12 @@ export type LDCApplicant = ApplicantBase & {
    * @description Information about the property owners, if different than the applicant
    */
   ownership:
-    | {interest: Extract<OwnersInterest, 'owner'>}
+    | {interest: Extract<OwnersInterest, 'owner'>} // sole owner does not report `owners`
+    | {
+        interest: Extract<OwnersInterest, 'other'>;
+        interestDescription: string; // "other" interest uniquely requires a free text description
+        owners: (OwnersNoticeGiven | OwnersNoNoticeGiven)[];
+      }
     | {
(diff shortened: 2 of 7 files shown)


examples/application/data/planningPermission/major.ts
```diff
@@ -46,6 +46,7 @@ export const planningPermissionMajor: Application = {
               country: 'Greece',
               postcode: '212 00',
             },
+            interest: 'Owner',
```
jessicamcinchak, line 49: This PP-Major example failed our automated tests because it was missing the new free-text `interest` (as will be required/collected via `OwnershipCertificateOwners` List Schema) which is good !

types/schemas/prototypeApplication/data/Applicant.ts
```diff
@@ -40,7 +40,12 @@ export type LDCApplicant = ApplicantBase & {
    * @description Information about the property owners, if different than the applicant
    */
   ownership:
-    | {interest: Extract<OwnersInterest, 'owner'>}
+    | {interest: Extract<OwnersInterest, 'owner'>} // sole owner does not report `owners`
+    | {
+        interest: Extract<OwnersInterest, 'other'>;
+        interestDescription: string; // "other" interest uniquely requires a free text description
+        owners: (OwnersNoticeGiven | OwnersNoNoticeGiven)[];
+      }
```
jessicamcinchak, line 48: The `prototypeApplication` is not actually exchanged with BOPS yet, but this is a nice example of how we're able to accomplish stronger types with it - we can specifically specify that when the interest is "other" within an LDC then there will be a freetext requirement "interestDescription".

Whereas for the current `application` schema below we simply introduce `interestDescription` as always optional (since PP "other" interest does _not_ require the freetext!)

types/shared/Ownership.ts
```diff
@@ -47,21 +48,24 @@ export type Owners = OwnersNoticeGiven | OwnersNoNoticeGiven | OwnersNoticeDate;
 
 export interface BaseOwners {
   name: string;
-  address: Address | string;
-  interest?: OwnersInterest;
+  address: Address;
```
jessicamcinchak, line 51: Tidying up this `Address | string` is a direct side effect of the List component now supporting "Address" fields (eg no long text for multiple owners paths!) :tada:

examples/application/data/planningPermission/major.ts
augustlindemer: Not sure I fully follow the above. Ownership certificates do not collect any _interest_ specification. The ownership certificates list only collects name, address, and dates. See: https://github.com/theopensystemslab/planx-new/pull/4248/files#diff-cb26b72aba100b736fd0d3c7193f4bf602baf84135f280677fdbe4ca4ae0b858.

Interest values for anyone but the applicant is collected only for LDC applications (for now, there are other app types that have the same/similar 'interest in land' sections) where applicant.ownership.interest = other. It's not a field in the form but we require the user to tell us in PlanX to avoid potential issues with validation.

jessicamcinchak: Excellent catch again - this an odd case where it's actually the example data that was wrong (eg this PP Major had `noticeGiven` _not_ `noticeDate` and was therefore expecting `interest` too!)

Good news is I think the typescript is right, and we just have some outdated examples haunting us - now corrected to use `noticeDate` _without_ `interest` !

augustlindemer (approved):
Looks good, let's test.

Issue #296, opened 2025-01-22
API implementation documentation


apricot13:
One of our aims is to make the DPR backend agnostic between councils as well as compliant with the ODP schema. The data side is pretty well handled however there is currently no official guidance on implementing an API, specifically for response structures, pagination, searching and filtering.

We would like to propose creating a set of **API implementation guidelines**, similar to the [Open Referral API Implementation Guidelines](https://developers.openreferraluk.org/API-Implementation/) to serve as a standard for all developers and ensure consistent implementation across councils.

> [!IMPORTANT]
> The below is now written up and merged into the main branch you can view the latest version [here](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/main/IMPLEMENTATION.md) 

# Proposed Documentation Topics:

1. **API response structure**
2. **Endpoints**
3. **Pagination**
4. **Searching and Filtering**

# API response structure

We propose a structure like the following to be used for all API responses. 

This is a slight change from how we're handling data ourselves currently but this approach feels more standardised and robust.

```ts
export interface ApiResponse<T> {
  data: T | null;
  pagination?: Pagination | CursorPagination;
  status?: {
    code: number;
    message: string;
    detail?: string;
  };
}
```

This structure includes:

- `data`: The actual data being returned.
- `pagination`: Optional pagination data, which could be either **offset** or **cursor-based**.
- `status`: Contains the HTTP status code, a message, and optional details. (see below for some proposals around this)

Inside the status section the `code` and `message` fields would mostly be used for standard error codes but the details section may become more specific over time and contain error messages related to pagination, filtering, or invalid query parameters. It would be good to have a place to align on the wording of these common errors. 

# Endpoints

DPR is closely tied with BOPs so our endpoint suggestions are mostly for public facing endpoints

- `GET /api/@next/public/planning_applications/search` 
  - for listing, searching and filtering planning applications that are currently available to the public
- `GET /api/@next/public/planning_applications/{reference}` 
  - returns the details of specific public planning application reference, we're hoping to propose a public schema for this data, hence the separate submission endpoint, although the redacted submission could also be appended to this endpoint too
- `GET /api/@next/public/planning_applications/{reference}/documents` 
  - for listing, searching and filtering documents related to the specific public planning application reference
- `GET /api/@next/public/planning_applications/{reference}/comments` 
  - for listing, searching and filtering comments related to the specific public planning application reference
[...]

PR #318, opened 2025-03-03
API Implementation Documentation

apricot13:
[Read it here](https://github.com/theopensystemslab/digital-planning-data-schemas/blob/hn/api-implementation-documentation/IMPLEMENTATION.md)!

This PR is to accompany the issue I raised here a while back (https://github.com/theopensystemslab/digital-planning-data-schemas/issues/296) I've stuck to just using markdown for now, following a "swagger-like" structure but I'm happy to switch to a different library or validator tool if there are any suggestions. 

I've based the documentation on:

* the endpoints we're currently using from BOPS
* the endpoints we will be changing in BOPS 
* The [DPR prototype](https://public-index-design-40bd5591d858.herokuapp.com/) designs which are currently being updated
* What I originally wrote in this issue: https://github.com/theopensystemslab/digital-planning-data-schemas/issues/296 

 
I've only covered the basics for now, we're starting work on comments first, then documents, then applications and then location based application searching. 

I've put in a TODO at the bottom of the implementation page to spark as much discussion as possible before we implement those in the DPR and add to the documentation.

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -1,0 +1,538 @@
+# API implementation guidelines
+
+One of the aims of this specification is to make data interopable between systems. JSON schema allows us to do this with the data structure but the implementation of API's will also need to have standard implementations in order for applications to use the data efficiently.
+
+## Formatting data
+
+##### Dates
+
+- The format for date only values is `YYYY-MM-DD`. Date keys should have the suffix `Date`.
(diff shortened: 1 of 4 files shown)


jessicamcinchak (approved):
Sorry have been slow to get to this one - thanks for clear guidelines & including this !

madz-lw (approved):

PR #322, opened 2025-03-06
Add AdvertConsent schema

christinelangston:
Edited the Applicant, Proposal, ApplicationData, Property types. Added extra filetype to accommodate AdvertConsent application.

### DIFF
--- a/types/schemas/prototypeApplication/data/Proposal.ts
+++ b/types/schemas/prototypeApplication/data/Proposal.ts
@@ -311,7 +311,6 @@ export type AdvertConsentProposal = Pick<ProposalBase, 'boundary'> & {
     projecting: Integer;
   };
   otherAdvertisements: boolean;
-  started: boolean;
   existingAdvertRemovedOrReplaced: boolean;
   highwayProjection: boolean;
   visibility: {public: boolean};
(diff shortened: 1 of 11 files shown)


types/schemas/prototypeApplication/data/Proposal.ts
```diff
@@ -300,6 +300,23 @@ export type HedgerowRemovalNoticeProposal = Pick<ProposalBase, 'boundary'> & {
...
+  advertType: string;
+  communityConsultation: boolean;
+  advertisementDetails: {
+    other: Integer;
+    fascia: Integer;
+    hoarding: Integer;
+    projecting: Integer;
+  };
+  otherAdvertisements: boolean;
+  started: boolean;
```
jessicamcinchak, line 314: nit: do we need `started: boolean;` if the `ProposalDates` above include the actual `start` date? Perhaps redundant and we can favor only dates?

jessicamcinchak (approved):
Looks great! One minor comment/suggestion :+1:

types/schemas/prototypeApplication/data/Proposal.ts
christinelangston: Good catch! I was wondering the utility of this as well. I'll update.

PR #337, opened 2025-04-25
refactor: remove `PrimaryApplicationType` in favor of granular application types only

jessicamcinchak:
Simplifying some conditional logic here, but need to touch a lot of files to do so! This change primarily impacts the `prototypeApplication` schema, which trickles down to the `postSubmission` schema (but only because it nests the original application data, does not change any add-ons like validation, consultations, etc).

Our idea of `TPrimary` "prefix" groupings isn't holding up as we test more services and we instead want to define specs per granular application type, but still drawing on as many shared type definitions as possible.

Key example is `ldc` prefix fails to correctly illustrate that `ldc.listedBuildingWorks` is _not_ a fee carrying service _without_ pre-assessment data, while all other granular types of LDC's do carry a fee.

Ability to define requirements by granular application types, rather than prefixed groups, will also more easily align to MHCLG's [application types](https://github.com/digital-land/planning-application-data-specification/discussions/categories/application-type).


examples/prototypeApplication/advertConsent.json
```diff
@@ -497,27 +472,7 @@
...
-            "one": []
-          },
-          "twelve": {
-            "one": [],
-            "two": []
-          },
-          "thirteen": [],
-          "fourteen": []
-        }
+        "payable": []
```
jessicamcinchak, line 475: Tidying up some application examples as I go to remove un-populated bits that are not relevant to this application type !

types/schemas/postSubmissionApplication/index.ts
```diff
@@ -1,57 +1,34 @@
...
  *
  * @todo create redacted form of PrototypeApplication
  */
-interface ApplicationSpecification<
-  TPrimary extends PrimaryApplicationType,
-  TGranular extends ApplicationType,
-> {
-  applicationType: TGranular;
+interface ApplicationSpecification<T extends ApplicationType> {
+  applicationType: T;
```
jessicamcinchak, line 22: This is the main change ! `T` is now always a granular `ApplicationType` enum value

types/schemas/prototypeApplication/data/ApplicationData.ts
```diff
@@ -58,25 +58,36 @@ export type PPApplicationData = FeeCarryingApplicationData & {
...
-  pa: FeeCarryingApplicationData;
-  pp: PPApplicationData;
-  listed: NonFeeCarryingApplicationData;
-  landDrainageConsent: FeeCarryingApplicationData;
-  wtt: NonFeeCarryingApplicationData;
-  hedgerowRemovalNotice: NonFeeCarryingApplicationData;
+  'ldc.breachOfCondition': FeeCarryingApplicationData;
+  'ldc.existing': FeeCarryingApplicationData;
+  'ldc.listedBuildingWorks': NonFeeCarryingApplicationData;
+  'ldc.proposed': FeeCarryingApplicationData;
```
jessicamcinchak, line 67: Per recent testing, we can now correctly encode which types of LDCs carry a fee & which don't :heavy_check_mark:

types/schemas/prototypeApplication/enums/ApplicationType.ts
```diff
@@ -587,9 +506,106 @@ type PPOutlineMajorSomeWaste = 'pp.outline.major.some.waste';
...
+ */
+type WTTNotice = 'wtt.notice';
+
+/**
+ * @id #ApplicationType
+ * @description Planning application types
+ */
+export type ApplicationType =
+  | AdvertConsent
+  | Amendment // Service descriptor, not actually submittable
```
jessicamcinchak, line 535: These `// Service descriptor, not actually submittable` comments are a little tricky ! 

We want to still include "prefixes" in overall list of enum values because these enable data visualisation like tree maps/dendrograms I think, but we don't actually ever expect a plain "ldc" to get submitted - hence example payloads will always be granular type like `ldc.existing`, etc ! 

May explore removing these later but should be harmless for now :ok_hand:

types/schemas/prototypeApplication/index.ts
```diff
@@ -4,81 +4,109 @@ import {ApplicationData} from './data/ApplicationData';
...
+type PriorApprovalPart20ClassAB =
+  ApplicationSpecificationWithPreAssessment<'pa.part20.classAB'>;
+type PlanningPermissionHouseholder =
+  ApplicationSpecification<'pp.full.householder'>;
+type PlanningPermissionHouseholderRetrospective =
+  ApplicationSpecification<'pp.full.householder.retro'>;
+type PlanningPermissionMajor = ApplicationSpecification<'pp.full.major'>;
+type PlanningPermissionMinor = ApplicationSpecification<'pp.full.minor'>;
+type WorksToTreesConsent = ApplicationSpecification<'wtt.consent'>;
+type WorksToTreesNotice = ApplicationSpecification<'wtt.notice'>;
```
jessicamcinchak, line 86: Note here that each granular variant is actually one of two options now: `ApplicationSpecification` or `ApplicationSpecificationWithPreAssessment`. This allows us to make outer `preAssessment` property requried or properly omit it, rather than always optional !

types/schemas/postSubmissionApplication/data/Assessment.ts
```diff
@@ -89,16 +89,20 @@ export type PriorApprovalAssessment = PostSubmissionAssessment & {
...
- * TypeMap of PrimaryApplicationTypes to their specific Assessment models
+ * TypeMap of granular application types to their specific Assessment models
  */
 type AssessmentVariants = {
-  pa: PriorApprovalAssessment;
+  'pa.part1.classA': PriorApprovalAssessment;
+  'pa.part3.classMA': PriorApprovalAssessment;
+  'pa.part7.classM': PriorApprovalAssessment;
+  'pa.part14.classJ': PriorApprovalAssessment;
+  'pa.part20.classAB': PriorApprovalAssessment;
```
DafyddLlyr, line 99: Great to see this being applied right away 👍

DafyddLlyr (approved):
Great! Both simpler and more accurate 👌

types/schemas/postSubmissionApplication/index.ts
```diff
@@ -61,74 +38,47 @@ interface ApplicationSpecification<
...
+  ApplicationSpecification<'pa.part14.classJ'>;
+export type PostSubmissionPAPart20ClassAB =
+  ApplicationSpecification<'pa.part20.classAB'>;
+export type PostSubmissionPPHouseholder =
+  ApplicationSpecification<'pp.full.householder'>;
+export type PostSubmissionPPMajor = ApplicationSpecification<'pp.full.major'>;
+export type PostSubmissionPPMinor = ApplicationSpecification<'pp.full.minor'>;
+export type PostSubmissionRightsOfWayOrder =
+  ApplicationSpecification<'rightsOfWayOrder'>;
+export type PostSubmissionWTT = ApplicationSpecification<'wtt'>;
```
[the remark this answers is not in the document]
jessicamcinchak: Yes ! Aware both this list and ours are actually still a subset and a bit incomplete - happy to take a pass over them in a followup PR to stub out all possible variants & ensure we're keeping naming consistent :+1:

types/schemas/prototypeApplication/enums/ApplicationType.ts
apricot13, line 1: We're not always so granular on the DPR side, we make use of the service descriptors in our help pages for example, https://planningregister.org/camden/help/application-types 

is it possible to still export those group titles as a sort of utility/helper function, what was previously known as `PrimaryApplicationType`? 🙏

https://github.com/tpximpact/digital-planning-register/blob/main/src/lib/planningApplication/type.tsx#L32

apricot13 (approved):
So much easier!

One downside though is that now theres an even longer list of outputs to generate! Thinking ahead to when this is an NPM package (😉) it would be super useful to have an object with all the latest application groups/types available. We're currently maintaining our own lists [here](https://github.com/tpximpact/digital-planning-register/blob/74e0a95a5fa44b14e9a3a7fecba0170388bcea22/src/lib/planningApplication/type.tsx#L32C44-L32C66) and [here](https://github.com/tpximpact/digital-planning-register/blob/main/src/lib/planningApplication/type.tsx#L341) for various purposes but if we can have an 'application type coverage checker' in the tests to make sure we haven't missed any out that would be helpful! 😂

types/schemas/prototypeApplication/enums/ApplicationType.ts
jessicamcinchak: Super helpful context, thanks ! Now still exporting `PrimaryApplicationType` type here https://github.com/theopensystemslab/digital-planning-data-schemas/pull/337/commits/87665d6a5ab44627ea31812172d10625de20d4fa :heavy_check_mark: 

NPM is still very much on our aspirational todo list - started some groundwork for this awhile back and hope to pick it back up in near future ! Will very much be keeping complimentary utility/helper function exports in mind then in addition to exported types :slightly_smiling_face:

PR #346, opened 2025-05-02
Update pagination structure to include totalResults and totalItems fields

apricot13:
Updates the pagination to clarify totalResults vs totalItems. I have set totalItems to be optional since its not always feasible to calculate this value.

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -60,9 +60,13 @@ export interface Pagination {
    */
   totalPages: number;
   /**
-   * Total number of items eg 100
+   * Represents the total number of items returned by the current query
    */
-  totalItems: number;
+  totalResults: number;
+  /**
+   * Represents the total number of items in the database (unfiltered)
+   */
+  totalItems?: number;
 }
 ```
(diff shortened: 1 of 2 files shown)


apricot13:
@benjamineskola  yeh I can kind of see what you mean, you have to think about it a beat too long each time you look at it. if you were give `totalResults` or `totalItems` out of context you could be confused if the numbers came up differently!

I've updated it to use `totalResults` and `totalAvailableItems` to help make the context a bit clearer when their seen in isolation.

**`totalResults`**
Using _results_ instead of items reinforces that the value reflects the outcome of a specific query or request - not the whole dataset. It's subtle, but it should help avoid any confusion.

**`totalAvailableItems`** 
Adding _available_ helps clarify that this number represents everything you could access if you paged through the full set of items in the dataset, not just what was returned in this response.

Hopefully this makes it a bit easier to quickly read or talk about: `totalResults` is what you’ve got in this response, and `totalAvailableItems` is the full set that’s available to query.

gabimartini (approved):

PR #354, opened 2025-05-26
feat: add a first draft `Enforcement` schema

jessicamcinchak:
A number of councils build & run "enforcement" services like "Report a planning breach" or "Report a derelict building" on PlanX and BOPS would like to accept these in the future. 

This PR introduces a starter `Enforcement` schema based on current services. Please not that as a discretionary service, there is not as much structured `data` available as statutory services - feedback about desired structured data should go to the RAB working group within ODP. The schema is built _from services_, not the other way around!

https://trello.com/c/msiSSiK9/3301-create-a-draft-of-an-enforcement-schema-with-a-rab-example-payload


examples/enforcement/breach.json
```diff
@@ -0,0 +1,397 @@
...
+    },
+    {
+      "question": "Describe the work suspected to be a breach",
+      "responses": [
+        {
+          "value": "Erection of a library in the rear garden"
+        }
+      ],
+      "metadata": {}
+    },
```
jessicamcinchak, line 304: "Select the breach you are reporting" and "Describe the work suspected to be a breach" seem like good candidates for `data.proposal.projectType` and `data.proposal.description` or similar, but are not currently built to use schema variables in the editor.

examples/enforcement/breach.json
```diff
@@ -0,0 +1,397 @@
...
+    {
+      "question": "Do you have any supporting documents or photos?",
+      "responses": [
+        {
+          "value": "No"
+        }
+      ],
+      "metadata": {}
```
jessicamcinchak, line 366: It's not required to upload any files - hences `files?` in the main type definition. If files are uploaded, they'll be labeled according to our usual types

DafyddLlyr (approved):

Issue #352, opened 2025-05-21
Windows-related issues


madz-lw:
When running the repo on my windows machine I have encountered a few issues.
First, running the `build-schema` script:
`"build-schema": "./scripts/build-schema.sh",`
When running this script, even in a bash terminal, I get issues with the ./ folder access convention. 

![Image](https://github.com/user-attachments/assets/ffcbc0fd-4258-4bbd-9f8e-c1f23ef627cc)

To resolve this manually, I prefixed the command with "bash":
`"build-schema": "bash ./scripts/build-schema.sh",`

Additionally, when running the build script itself (which needed the above change), it generates a lot of example json files:

![Image](https://github.com/user-attachments/assets/d0f91213-2816-467d-b6d2-1d1e4d679506)

I don't believe these should all be generated so maybe there is another windows-related caveat causing this?

PR #361, opened 2025-06-04
#352 Run build schema script in node

apricot13:
If at first you don't succeed, write it in typescript....suggested fix for #352 🤞

I'm not able to test as I don't have windows machine handy


apricot13:
This issues remain I'm told?

> example .json files are writing to wrong place - eg examples/application/data/landDrainageConsent.json should be examples/application/landDrainageConsent.json (only TS files within /data subdirectory)

madz-lw:
image

Looks fine for me. No .json files in /data

